PR #5, opened 2024-09-21
rewrite

peer2f00l:


src/vault.rs
```diff
@@ -1,61 +1,349 @@
...
-                deposits: Vec::new(),
-                min_collateral_ratio,
-                debt: 0,
-            },
-        )
+        prefix: impl IntoStorageKey,
+        loan_asset_id: FungibleAsset,
+        collateral_asset_id: FungibleAsset,
+        min_collateral_ratio: (u8, u8),
+        nft_gate: Option<AccountId>,
```
royalf00l, line 66: again, gating is required, so >= 1 NFT collection identifiers

src/vault.rs
```diff
@@ -1,61 +1,349 @@
...
+    }
+
+    pub fn record_liquidation(&mut self, borrower_id: &AccountId) {
+        let mut borrower_entry = self.borrowers.get(borrower_id).unwrap_or_default();
+
+        let liquidated_collateral = borrower_entry.collateral_asset_deposited;
+        borrower_entry.collateral_asset_deposited = 0;
+        let liquidated_loan = borrower_entry.loan_asset_borrowed;
+        borrower_entry.loan_asset_borrowed = 0;
+        // TODO: Do we distribute the liquidated collateral as fees/rewards?
```
royalf00l, line 332: See Figure 1 in the [litepaper](https://docs.google.com/document/d/1vugo4tqSoOdqHUZlPkFxN9s1wbRpsnCQ/edit#heading=h.gjdgxs) for how the spread is distributed. To start, we can simply have a protocol, liquidator, and lender split which is set in the vault

src/vault.rs
```diff
@@ -1,61 +1,349 @@
...
+        // TODO: Do we swap the liqidated funds to the loan asset?
```
royalf00l, line 333: The liquidator will be selling the collateral asset into the loan asset. They will take care of that for us and then collect the liquidator reward. We need to think through how the liquidation process occurs in detail to prevent the liquidators from keeping all the collateral assets. There should be some sort of liquidation function that handles this

royalf00l (changes requested):
Thanks for the rewrite. Couple small changes requested, but overall makes sense

src/lib.rs
```diff
@@ -1,350 +1,101 @@
...
-
-        self.nft.nft_token(token_id).unwrap()
-    }
+#[near(serializers = [json])]
+pub struct ViewVault {
+    pub vault_id: U64,
+    pub loan_asset_id: FungibleAsset,
+    pub collateral_asset_id: FungibleAsset,
+    pub min_collateral_ratio: (u8, u8),
+    pub nft_gate: Option<AccountId>,
```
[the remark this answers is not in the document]
royalf00l: I considered the scenario where the user transfers the NFT after taking out a loan and that loan would then be attached to the NFT afterwards. Same goes for a lender. If a lender loans out stables, then transfers the NFT, the new NFT owner can now redeem the loan and collect the yield (After the minimum amount of time has passed). So in general, the NFT would still be required for performing major actions (borrow, lend, repay, etc).

I think what I described is closer to the latter flow you described. Lmk if you agree.

src/vault.rs
```diff
@@ -1,61 +1,349 @@
...
+use crate::asset::FungibleAsset;
+
+#[derive(BorshStorageKey)]
+#[near]
+enum StorageKey {
+    NftGate,
+    Lenders,
+    Borrowers,
+    TotalLoanAssetDepositedLog,
+    CollateralAssetFeeDistributionLog,
```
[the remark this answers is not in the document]
royalf00l: Yes. That's the job of the liquidator. I'm not decided on how specifically that's implemented. Do you think that the protocol itself acting as a liquidator to start is OK? Where we sell on ref.finance for example?

src/vault.rs
royalf00l: I'm open to the contract being the liquidator to start, but if you have another idea that you think would work better, I'm open to that too.

src/vault.rs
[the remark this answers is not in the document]
royalf00l: I'm thinking if the liquidator makes a call to liquidate passing in a liquidation venue parameter, then the contract looks at the pyth oracle price (XCC w/ gas paid for by liquidator), calculates the Collateral Ratio & compares to MCR, then if the CR < MCR the contract liquidates on the venue.

`fn liquidate(vault_id: u64, lender_account_id: AccountId , liq_venue: String)`

royalf00l (approved):
Let's merge these changes into main and deploy on testnet

PR #6, opened 2025-01-08
Dev/fresh start

peer2f00l:


common/src/withdrawal_queue.rs
```diff
@@ -92,7 +101,35 @@ impl WithdrawalQueue {
         }
     }
 
-    pub fn pop(&mut self) -> Option<(AccountId, u128)> {
+    pub fn try_lock(&mut self) -> Option<(AccountId, BorrowAssetAmount)> {
```
royalf00l: I'm not an expert on NEAR's contract runtime, but in the `try_lock` function, there might be a potential race condition between checking `is_locked` and setting it. In a concurrent environment, multiple transactions could pass the check before one sets the lock.

common/src/market/configuration.rs
```diff
@@ -28,13 +25,20 @@ pub struct MarketConfiguration {
...
+    /// How far below market rate to accept liquidation? This is effectively the liquidator's spread.
+    ///
+    /// For example, if a 100USDC borrow is (under)collateralized with $110 of
+    /// NEAR, a "maximum liquidator spread" of 10% would mean that a liquidator
+    /// could liquidate this borrow by sending 109USDC, netting the liquidator
+    /// ($110 - $100) * 10% = $1 of NEAR.
+    pub maximum_liquidator_spread: Rational<u16>,
 }
 
 impl MarketConfiguration {
```
royalf00l, line 45: The `MarketConfiguration` struct lacks validation for its fields. For example, `maximum_liquidator_spread` should probably be less than 1, and `minimum_borrow_amount` should be greater than zero. We may want to consider adding validation.

common/src/withdrawal_queue.rs
```diff
@@ -199,12 +275,29 @@ impl<'a> Iterator for WithdrawalQueueIter<'a> {
...
+    pub length: u32,
+}
+
 #[cfg(test)]
 mod tests {
     use near_sdk::AccountId;
 
     use super::WithdrawalQueue;
 
+    // TODO: Test locking.
```
royalf00l, line 347: I see you're already aware of potential withdrawal queue concurrency issues

common/src/market/impl.rs
```diff
@@ -55,216 +54,226 @@ impl Market {
...
             )),
             withdrawal_queue: WithdrawalQueue::new(key!(WithdrawalQueue)),
+            static_yield: LookupMap::new(key!(StaticYield)),
         }
     }
 
-    pub fn get_borrow_position(&self, account_id: &AccountId) -> Option<BorrowPosition> {
-        self.borrow_positions.get(account_id)
-    }
+    pub fn try_lock_next_withdrawal_request(
```
royalf00l, line 93: The withdrawal request process involves multiple state changes. If any step fails, there's no rollback mechanism.

common/src/market/mod.rs
```diff
@@ -19,20 +22,24 @@ pub use r#impl::*;
...
     /// Available to be borrowed right now.
-    pub available: U128,
-    pub deposited: U128,
+    pub available: BorrowAssetAmount,
+    pub deposited: BorrowAssetAmount,
 }
 
 impl BorrowAssetMetrics {
-    pub fn calculate(deposited: u128, balance: u128, maximum_usage_ratio: Rational<u128>) -> Self {
+    pub fn calculate(
```
royalf00l: The subtraction and multiplication operations in this fn could potentially overflow. While there's a `checked_scalar_mul`, other operations should also use checked arithmetic.

common/src/withdrawal_queue.rs
```diff
@@ -92,7 +101,35 @@ impl WithdrawalQueue {
...
+    pub fn unlock(&mut self) {
+        self.is_locked = false;
+    }
+
+    /// Only pops if:
+    /// 1. Queue is non-empty.
+    /// 2. Queue is locked.
+    ///
+    /// Unlocks the queue.
+    pub fn try_pop(&mut self) -> Option<(AccountId, BorrowAssetAmount)> {
```
royalf00l, line 136: If `try_pop` panics after setting `is_locked = false` but before completing the operation, the queue could be left in an inconsistent state. Maybe consider using a RAII-style lock guard pattern. 
https://aloso.github.io/2021/03/18/raii-guards.html
https://effective-rust.com/raii.html

common/src/market/impl.rs
```diff
@@ -55,216 +54,226 @@ impl Market {
...
+    fn log_borrow_asset_deposited(&mut self, amount: BorrowAssetAmount) {
         let block_height = env::block_height();
         self.total_borrow_asset_deposited_log
             .insert(&block_height, &amount);
     }
 
-    fn record_borrow_asset_reward_distribution(&mut self, amount: u128) {
+    fn record_borrow_asset_yield_distribution(&mut self, mut amount: BorrowAssetAmount) {
+        // First, static yield.
+        // TODO: Lots of `.unwrap()`s in this section.
```
royalf00l: Multiple unwrap calls in yield distribution could panic. The comment "TODO: Lots of `.unwrap()`s in this section" acknowledges this. These should be properly handled with Result types.

common/src/market/impl.rs
```diff
@@ -310,31 +324,19 @@ impl Market {
...
-            borrow_position.zero_out_collateral_asset_deposit();
-
-        // TODO: bounds checks
-        self.collateral_asset_balance -= collateral_asset_amount_liquidated;
-        self.borrow_asset_balance += recovered_borrow_asset_amount;
-
-        if let Some(margin) =
-            recovered_borrow_asset_amount.checked_sub(borrow_position.borrow_asset_liability.0)
+        if recovered_amount
+            .split(borrow_position.get_borrow_asset_principal())
```
royalf00l: I'd consider a loss anything that doesn't recover the principal + interest, but we can just put that down as a TODO for later.

src/lib.rs
```diff
@@ -63,61 +69,86 @@ impl FungibleTokenReceiver for Contract {
...
-                    "This market does not support collateralization with this asset",
-                );
+                let amount = use_collateral_asset();
+
+                let mut borrow_position = self
+                    .borrow_positions
+                    .get(&sender_id)
+                    .unwrap_or_else(|| BorrowPosition::new(env::block_height()));
 
                 // TODO: This creates a borrow record implicitly. If we
```
royalf00l: The sign-up step would only be NFT gating or something of that sort, which is just an additional pre condition check.

src/lib.rs
```diff
@@ -63,61 +69,86 @@ impl FungibleTokenReceiver for Contract {
...
-
-                // TODO: This function *errors* on overpayment. Instead, add a
-                // check before and only repay the maximum, then return the excess.
-                self.record_borrow_position_borrow_asset_repay(&sender_id, amount.0);
-
-                PromiseOrValue::Value(U128(0))
+                let amount = use_borrow_asset();
+
+                if let Some(mut borrow_position) = self.borrow_positions.get(&sender_id) {
+                    // TODO: This function *errors* on overpayment. Instead, add a
```
royalf00l: Due to the slightly imprecise calculation of yield and other fees, the returning of the excess should be anything >1%, for example, over the total amount borrowed + fees/interest.

src/lib.rs
```diff
@@ -128,7 +159,9 @@ impl FungibleTokenReceiver for Contract {
 
                 // TODO: Do we need to check the value of the amount recovered?
```
royalf00l: The amount recovered post liquidation should be in stablecoins. The only reason we'd need to check the amount recovered is in the event of a stablecoin depeg (let's not worry about that for now).

src/lib.rs
```diff
@@ -216,72 +245,293 @@ impl MarketExternalInterface for Contract {
...
-        PromiseOrValue::Promise(
-            self.configuration
-                .borrow_asset
-                .transfer(env::predecessor_account_id(), amount.0),
-        )
+        self.borrow_positions.insert(&account_id, &borrow_position);
+
+        self.configuration
+            .borrow_asset
+            .transfer(account_id, amount) // TODO: Check for failure
```
royalf00l: important TODO

src/lib.rs
```diff
@@ -216,72 +245,293 @@ impl MarketExternalInterface for Contract {
...
+                ),
+                "Borrow must still be above MCR after collateral withdrawal.",
+            )
+        }
+
+        self.borrow_positions.insert(&account_id, &borrow_position);
+
+        self.configuration
+            .collateral_asset
+            .transfer(account_id, amount) // TODO: Check for failure
```
royalf00l: important TODO

src/lib.rs
```diff
@@ -216,72 +245,293 @@ impl MarketExternalInterface for Contract {
...
+            )
+        } else {
+            None
+        };
+
+        match (borrow_promise, collateral_promise) {
+            (Some(b), Some(c)) => b.and(c),
+            (Some(p), _) | (_, Some(p)) => p,
+            _ => env::panic_str("No yield to withdraw"),
+        } // TODO: Check for success
```
royalf00l: Important TODO

src/lib.rs
```diff
@@ -216,72 +245,293 @@ impl MarketExternalInterface for Contract {
...
+    #[private]
+    pub fn after_execute_next_withdrawal(&mut self, account: AccountId, amount: BorrowAssetAmount) {
+        // TODO: Is this check even necessary in a #[private] function?
+        require!(env::promise_results_count() == 1);
+
+        match env::promise_result(0) {
+            PromiseResult::Successful(_) => {
+                // Withdrawal succeeded: remove the withdrawal request from the queue.
+
+                // TODO: If this panics, this is BIG BAD, as it means there is
```
royalf00l: important TODO

royalf00l (changes requested):
Overall looks good. There are some edge cases that could be problematic (withdrawal queue, config validation, unsafe arithmetic (overflow potential), and some unwraps). 

I think if you clean up the withdrawal queue, config validation, and important TODOs that auditors should have much more helpful feedback on arithmetic and unwraps than I can provide.

common/src/withdrawal_queue.rs
peer2f00l: Receipts that interact with a particular contract are guaranteed to execute sequentially, even within the same block. The `is_locked` flag is intended to prevent cross-block race conditions (e.g. between attempting and potentially failing a withdrawal). For example, if the queue looks like:

1. Alice: $100
2. Bob: $10

But the contract only has $10 at the moment, the `is_locked` flag should prevent the following case:

1. Bob calls `execute_next_supply_withdrawal_request`, which pops Alice's request and attempts to send her $100, which will fail.
2. However, before it fails, Bob calls `execute_next_supply_withdrawal_request` *again*, popping Bob's request and attempting to send him $10.
4. Alice's withdrawal request fails, so her entry is re-added to the top of the queue.
5. Bob's request succeeds, so he has essentially "jumped the line" and gotten his withdrawal before his turn.

common/src/withdrawal_queue.rs
peer2f00l: A panic will revert all of the state changes that were attempted in the receipt.

RAII guards like this are great if your application stays running while the resource is "acquired," however, the resource acquisition here (which is something like "the right to have your withdrawal request be processed") spans multiple blocks, so we have to save the state of the resource acquisition in the contract state somewhere (e.g. the `is_locked` flag).

Currently, `try_pop` is only invoked in two places:
- https://github.com/Templar-Protocol/contract-mvp/blob/bbcf6e700f7f224553ded593194f3e9a9854a641/src/lib.rs#L508
    This is only invoked after [locking the queue](https://github.com/Templar-Protocol/contract-mvp/blob/bbcf6e700f7f224553ded593194f3e9a9854a641/src/lib.rs#L348)
- https://github.com/Templar-Protocol/contract-mvp/blob/bbcf6e700f7f224553ded593194f3e9a9854a641/common/src/market/impl.rs#L93
    This is also only invoked after [locking the queue](https://github.com/Templar-Protocol/contract-mvp/blob/bbcf6e700f7f224553ded593194f3e9a9854a641/common/src/market/impl.rs#L72)

common/src/market/configuration.rs
peer2f00l: Borrow amount must be nonzero, [enforced during borrow execution](https://github.com/Templar-Protocol/contract-mvp/blob/bbcf6e700f7f224553ded593194f3e9a9854a641/src/lib.rs#L251). I can add validation for this field at configuration time as well.

Is it definitely wrong to allow a >100% `maximum_liquidator_spread`? All this would mean is that, during liquidation, the market is *always* willing to accept a loss. Is this a configuration that we know will never be desirable? Maybe some sort of extremely low-MCR, extremely high-APY/fees market config, such that accepting a loss at liquidation is never an issue?

common/src/market/impl.rs
peer2f00l: There is this rollback:

https://github.com/Templar-Protocol/contract-mvp/blob/bbcf6e700f7f224553ded593194f3e9a9854a641/src/lib.rs#L528-L534

common/src/market/mod.rs
peer2f00l: The `require` will panic on `deposited < balance`. All other arithmetic operations in this function are checked. This function is only called in a view context.

src/lib.rs
peer2f00l: branch is not really in a "done" state, I guess the review is more for "did I fail to mark important spots with TODOs" or something like that

these will all be addressed in future issues

common/src/market/configuration.rs
royalf00l: I don't think the borrow amount itself needs to be validated in the configuration (it should be enforced at execution), but the `minimum_borrow_amount` should be (ie > 1 unit).

Is it definitely wrong to allow a >100% maximum_liquidator_spread?  - yes. The liquidation spread is what's left over after fees, interest, principal have been repaid, so it's possible that that the liquidation spread is 0 (It should be clipped at 0 and never calculated as negative). The market is always willing to accept a loss on the spread. Under no circumstances should the % of the total liquidation spread be >100%.  

Is this a configuration that we know will never be desirable? Maybe some sort of extremely low-MCR, extremely high-APY/fees market config, ... - yes, reasonable, but permissive boundary checks. These details can be voted on by DAO later.

test-utils/src/lib.rs
```diff
@@ -224,6 +234,33 @@ impl TestController {
...
 
+    pub async fn asset_transfer_call(
+        &self,
+        asset_id: &AccountId,
+        sender: &Account,
+        receiver_id: &AccountId,
+        amount: u128,
+        msg: &str,
+    ) -> ExecutionSuccess {
+        println!(
```
royalf00l, line 312: should we use log emission instead of prints?

peer2f00l: This is in a test controller, not in the contract, so this code is only executed (or even compiled) during tests.

SmolinPavel:
Closing this one to reopen a new one later
(discussed with @peer2f00l )

SmolinPavel:
A new PR with the same branches involved -> https://github.com/Templar-Protocol/contract-mvp/pull/32

PR #46, opened 2025-03-13
Supply withdrawal fee

### LINKED ISSUE #8: Enforce various configuration options
Currently many different configuration options exist but are unimplemented.

Ordered by priority, high to low.

- [x] `maximum_borrow_asset_usage_ratio` #12 
- [x] `maximum_borrow_duration_ms` #17 
- [x] `minimum_borrow_amount` #18 
- [x] `maximum_borrow_amount` #18

### LINKED ISSUE #41: `supply_withdrawal_fee`

peer2f00l:
Depends on #16

### DIFF
--- a/common/src/fee.rs
+++ b/common/src/fee.rs
@@ -54,7 +54,11 @@ pub enum TimeBasedFeeFunction {
 }
 
 impl<T: AssetClass> TimeBasedFee<T> {
-    pub fn of(&self, amount: FungibleAssetAmount<T>, time: u64) -> Option<FungibleAssetAmount<T>> {
+    pub fn of(
+        &self,
+        amount: FungibleAssetAmount<T>,
+        duration: u64,
+    ) -> Option<FungibleAssetAmount<T>> {
         let base_fee = self.fee.of(amount)?;
 
         if self.duration.0 == 0 {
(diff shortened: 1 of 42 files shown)


common/src/fee.rs
```diff
@@ -54,28 +54,41 @@ pub enum TimeBasedFeeFunction {
...
             TimeBasedFeeFunction::Logarithmic => Some(
                 // TODO: Seems jank.
                 #[allow(
                     clippy::cast_sign_loss,
                     clippy::cast_possible_truncation,
                     clippy::cast_precision_loss
                 )]
-                (((base_fee.as_u128() as f64 * f64::log2((1 + time - self.duration.0) as f64))
-                    / f64::log2((1 + time) as f64))
+                (((base_fee.as_u128() as f64 * f64::log2((1 + duration - self.duration.0) as f64))
```
royalf00l: Is there a reason we're having base_fee be a u128 dtype if we're converting it to an f64? If base fee can only fit as a u128, then couldn't there be scenarios in which the conversion to an f64 overflows like when the u128 is > 2^53 - 1? I'm not as concerned about a bit of precision loss, but if I'm correct about the overflow potential, that's concerning

royalf00l (commented):
Only concern is the u128 -> f64 conversion. Everything else lgtm

common/src/fee.rs
peer2f00l: I'm not sure if we'll actually use this fee strategy anyways, so I might as well remove it.

royalf00l (approved):

PR #54, opened 2025-03-17
Quality-of-life improvements

peer2f00l:
Mostly code cleanup.

Some public API changes to configuration:

- `minimum_initial_collateral_ratio` &rarr; `borrow_mcr_initial`
- `minimum_collateral_ratio_per_borrow` &rarr; `borrow_mcr`
- `maximum_borrow_asset_usage_ratio` &rarr; `borrow_asset_maximum_usage_ratio`
- `maximum_borrow_duration_ms` &rarr; `borrow_maximum_duration_ms`
- `minimum_borrow_amount` &rarr; `borrow_minimum_amount`
- `maximum_borrow_amount` &rarr; `borrow_maximum_amount`
- `maximum_liquidator_spread` &rarr; `liquidation_maximum_spread`


common/src/asset.rs
```diff
@@ -168,10 +170,14 @@ impl<T: AssetClass> FungibleAssetAmount<T> {
...
 
-    pub fn as_u128(&self) -> u128 {
+    pub fn to_u128(&self) -> u128 {
         self.amount.0
     }
 
+    pub fn to_decimal(&self) -> Decimal {
+        self.amount.0.into()
+    }
+
```
petarvujovic98, line 180: I usually approach these with Rust's `From` impls, something like:
```rust
impl<T: AssetClass> From<FungibleAssetAmount<T>> for u128 {
    fn from(value: FungibleAssetAmount<T>) -> Self {
        value.amount.0
    }
}
```

common/src/borrow.rs
```diff
@@ -203,14 +204,13 @@ impl<M: Borrow<Market>> LinkedBorrowPosition<M> {
...
-        let interest = interest_rate_part
-            * Decimal::from(self.position.get_borrow_asset_principal().as_u128());
+        let interest = interest_rate_part * self.position.get_borrow_asset_principal().to_decimal();
 
         interest.to_u128_ceil().unwrap().into()
     }
 
     pub(crate) fn calculate_interest(&self, limit: u32) -> AccumulationRecord<BorrowAsset> {
-        let principal = Decimal::from(self.position.get_borrow_asset_principal().as_u128());
+        let principal = self.position.get_borrow_asset_principal().to_decimal();
```
petarvujovic98, line 213: If this had a `From<...> for Decimal` impl block it could be replaced with:
```suggestion
        let principal = self.position.get_borrow_asset_principal().into();
```
which might not be as explicit for some folk, but I find it more idiomatic.

common/src/market/balance_oracle_configuration.rs
```diff
@@ -150,17 +151,8 @@ impl AssetConversion<CollateralAsset, BorrowAsset> for PricePair {
...
-            / (u128::from(self.borrow_asset_price.price)
-                - u128::from(self.borrow_asset_price.confidence))
-            * Decimal::from(10u32).pow(
-                self.collateral_asset_price.exponent_10
-                    - self.collateral_asset_decimals
-                    - self.borrow_asset_price.exponent_10
-                    + self.borrow_asset_decimals,
-            ))
+        (amount.to_decimal() * self.collateral_asset_price.upper_bound()
+            / self.borrow_asset_price.lower_bound())
```
petarvujovic98, line 155: Very nice

petarvujovic98 (approved):
Some small suggestions I left, but looks good overall.
One thing I'd suggest moving forward would be a tool like [Graphite](https://graphite.dev) to keep these changesets more reviewable.
This PR is touching a few things at the same time and thus a bit more invloved for the reviewer - at least more than could be if the changes would be devided in a few small PRs.
For example:

- Rename PR - just deals with the renaming of some fields since most of the files fall into this category, it is then trivial to review in 2 mins 
- Type change - moving to `Decimal` for certain fields or removing certain pairs in favor of single compound field
- Util/supply removal - again, simple change trivial to review

This still isn't all that difficult to review, but could be easier if separated in a similar approach to the above comment.

common/src/asset.rs
peer2f00l: I was originally trying to avoid this, since I wanted to be careful to avoid mixing values that come from asset-types with different `AssetClass` discriminants without being extremely explicit. However, this would still require a decently-long method chain, so maybe it would still be explicit enough. What do you think?

common/src/market/balance_oracle_configuration.rs
peer2f00l: There might actually be a downside with this change - the way it was previously would have preserved more fractional digits for longer. The new version truncates them sooner, so it means that the new version is definitely less-precise. However, I _think_ I should be able to do this just fine still, since we're immediately converting to an integer type. Agree/disagree?

petarvujovic98: I also think it should be fine. The level of precision loss seems acceptable for this

common/src/asset.rs
petarvujovic98: To my eyes it seems like the right approach

peer2f00l:
what, why did it merge?!

peer2f00l:
I'm turning off auto-merge.

common/src/borrow.rs
peer2f00l: https://github.com/Templar-Protocol/contract-mvp/pull/59

PR #73, opened 2025-03-21
Proper accounting of untracked "loose" funds

### LINKED ISSUE #64: Contract will allow assets obtained through unknown means to be borrowed, but then panics if usage is >100%
Scenario:

1. 100 USDT have been supplied to the contract.
2. Someone sends the contract 100 USDT "just because." This could be malicious, or accidental.
3. The contract sends out 150 USDT in borrows.
4. Now the contract has a usage ratio of 150%, which is (currently) interpreted as an invariant violation and induces a panic.

Two possible mitigations:
1. Do not allow "funds of unknown origin" to be used in borrows. This means that any funds controlled by the contract, but not tracked by the contract, are as-good-as burned.
2. Do allow "funds of unknown origin" to be used in borrows. This means that suppliers could see a "better than expected" return, since yield generated by those loose funds will still only be distributed to known suppliers. However, this could also lead to a higher-than-expected interest rate for borrowers, since the usage rate will reach 100% before the contract has exhausted all of the funds it can use for borrowing.

peer2f00l:
API interface changes:
- `get_borrow_asset_metrics` no longer takes `borrow_asset_balance` parameter, since it will use internal balance accounting for this instead.


common/src/market/external.rs
```diff
@@ -24,10 +24,7 @@ pub trait MarketExternalInterface {
     /// Takes current balance as an argument so that it can be called as view.
     /// `borrow_asset_balance` should be retrieved from the borrow asset
     /// contract specified in the market configuration.
```
petarvujovic98: Comment needs updating

common/src/snapshot.rs
```diff
@@ -14,8 +14,10 @@ pub struct Snapshot {
 
 impl Snapshot {
     pub fn usage_ratio(&self) -> Decimal {
-        if self.deposited.is_zero() {
+        if self.deposited.is_zero() || self.borrowed.is_zero() {
             Decimal::ZERO
+        } else if self.borrowed >= self.deposited {
+            Decimal::ONE
         } else {
             self.borrowed.to_decimal() / self.deposited.to_decimal()
```
petarvujovic98, line 22: Wouldn't it just be easier to return something like 
```rust
(self.borrowed.to_decimal() / self.deposited.to_decimal()).min(Decimal::ONE)
```

contract/market/tests/untracked_funds.rs
```diff
@@ -0,0 +1,41 @@
+use test_utils::*;
```
petarvujovic98, line 1: Just a nit, I usually prefer having all imports explicitly stated rather than having a `*` for everything. In this case it isn't all that impactful, but it can cause naming conflicts which are not immediately obvious due to the import statement.
Again, nothing major, just something that's been easier on me in the past.

petarvujovic98 (approved):
Looks good overall, just a small couple nits which are not blockers.

common/src/snapshot.rs
peer2f00l: Yes, but `Decimal` computations are relatively expensive, especially multiplication/division. While the suggestion is more concise I think it will probably cost more gas.

common/src/market/external.rs
peer2f00l: Good catch!

contract/market/tests/untracked_funds.rs
peer2f00l: Will be addressed in https://github.com/Templar-Protocol/contract-mvp/issues/72

PR #78, opened 2025-03-24
CI/CD caching & proper Rust toolchain setup

peer2f00l:


.github/workflows/deploy-staging.yml
```diff
@@ -22,9 +22,24 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@1.85.0
```
petarvujovic98, line 26: There's `1.85.1` now which fixes a small doctest issue - not really relevant to us, but noting it here

petarvujovic98 (approved):
Looks fine.

.github/workflows/deploy-staging.yml
peer2f00l: Good note; repo version was already set to 1.85.0, so this just ensures that the installed version is the same that is specified in `rust-toolchain.toml`.

PR #79, opened 2025-03-24
Resolve TODOs and Clippy lints

### LINKED ISSUE #72: Resolve all TODOs and `clippy::pedantic` lints
- Add clippy lint for number spacing with `_`.
- Add clippy lint for warning against glob imports.

peer2f00l:


common/src/borrow.rs
```diff
@@ -223,6 +223,8 @@ impl<M: Borrow<Market>> LinkedBorrowPosition<M> {
         let interest =
             interest_rate_part * Decimal::from(self.position.get_borrow_asset_principal());
 
+        // Assume interest will never exceed u128::MAX
+        #[allow(clippy::unwrap_used)]
```
petarvujovic98: Wouldn't writing a `reason` be more idiomatic:
```suggestion
        #[allow(clippy::unwrap_used, reason = "Assume interest will never exceed u128::MAX")]
```
P.S. fmt might fold into two lines

petarvujovic98 (approved):
Nice cleanup!

common/src/borrow.rs
peer2f00l: Okay, added a bunch. Thanks for this, I didn't know about this feature!

PR #82, opened 2025-03-24
Check for success on static yield withdrawal

peer2f00l:


contract/market/src/impl_helper.rs
```diff
@@ -378,4 +378,40 @@ impl Contract {
...
+            if matches!(env::promise_result(i), PromiseResult::Failed) {
+                static_yield
+                    .borrow_asset
+                    .join(borrow_asset_amount)
+                    .unwrap_or_else(|| {
+                        env::panic_str("Borrow asset static yield returned overflows")
+                    });
+            }
+            i += 1;
+        }
```
petarvujovic98, line 404: This doesn't have to be `mut`. Something like the following could be nicer:
```suggestion
        let i = if borrow_asset_amount.is_zero() {
            0
        } else {
            if matches!(env::promise_result(i), PromiseResult::Failed) {
                static_yield
                    .borrow_asset
                    .join(borrow_asset_amount)
                    .unwrap_or_else(|| {
                        env::panic_str("Borrow asset static yield returned overflows")
                    });
            }
            1
        };
```

petarvujovic98 (approved):

PR #83, opened 2025-03-24
Introduce the `self_ext!()` macro to reduce duplication

peer2f00l:
Because `Self::ext(env::current_account_id())` is long and difficult to read.


petarvujovic98 (approved):
Other than the tests failing, this seems like a nice QOL improvement

PR #84, opened 2025-03-24
Removes ability to withdraw supply yield directly

### LINKED ISSUE #81: Remove ability to withdraw yield directly
It should always be added to the main principal first so that fees can be simply applied and calculated.

peer2f00l:
We want to remove this for a few reasons:
1. The behavior can be closely approximated with a sequence of calls:
    - `harvest_yield(compounding = true)`
    - `create_supply_withdrawal_request(amount = <yield_amount>)`
    - `execute_next_supply_withdrawal_request()`
2. Allowing this behavior allows for a (small) bypass of the supply withdrawal fees & queue.
3. Reduces code & complexity.


petarvujovic98 (approved):
Nothing to comment on here

PR #99, opened 2025-03-27
Adds `snapshot_limit` optional parameter to `harvest_yield` function

### LINKED ISSUE #94: Allow partial computation of interest/yield via explicit calls
Either by dynamic gas checking or by allowing callers to specify a certain number of iterations, or by introducing a static limit.

This will allow the calculation to be spread out across multiple function calls, and avoid positions getting softlocked from running out of gas during interest/yield calculations.

peer2f00l:
This parameter allows the caller to specify an upper limit for the number of snapshots to process during the function call. This is useful in cases when the normal `harvest_yield` call might run out of gas.

This PR also adds this estimation calculation to the gas report generated as part of the CI/CD job.

# For UI

For now, it is probably easiest to make the "Harvest yield" button perform these two actions:

1. Call `harvest_yield(snapshot_limit: ...)` with `snapshot_limit` set to the value generated in the report below.
2. Call `harvest_yield(compounding: true)`.

### DIFF
--- a/common/src/borrow.rs
+++ b/common/src/borrow.rs
@@ -226,7 +226,10 @@ impl<M: Deref<Target = Market>> LinkedBorrowPosition<M> {
         interest.to_u128_ceil().unwrap().into()
     }
 
-    pub(crate) fn calculate_interest(&self, limit: u32) -> AccumulationRecord<BorrowAsset> {
+    pub(crate) fn calculate_interest(
+        &self,
+        snapshot_limit: u32,
+    ) -> AccumulationRecord<BorrowAsset> {
         let principal: Decimal = self.position.get_borrow_asset_principal().into();
         let mut next_snapshot_index = self.position.borrow_asset_fees.get_next_snapshot_index();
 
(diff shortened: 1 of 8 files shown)


peer2f00l:
Also do the same for `accumulate_interest`

SmolinPavel (approved):
Just wondering, is there a chance to merge 2 calls (snapshot_limit & compounding) in 1 call?

peer2f00l:
@SmolinPavel Possibly, but with some difficulty. Actually, _most of the time_, the UI should just be able to do the second `compounding: true` call. However, there is a potential failure case in which the contract runs out of gas and cannot complete it.

The full algorithm looks like this:

1. Get the most recent snapshot index `snapshot_height = snapshots.len()`. I can add a view function to easily retrieve this value.
2. Get the last snapshot when the supply record was updated `last_update = supply_position.borrow_asset_yield.next_snapshot_index`.
3. Check if the index delta `delta = snapshot_height - last_update` between those two is greater than the snapshot limit. That is:
    - For supply records, check if the delta is greater than 23,000 (approximately, see the "Gas report" generated by the CI bot).
    - If the delta is greater, then call the corresponding accumulation function (`harvest_yield`) with `snapshot_limit` set to the corresponding limit, and go to step 2.
4. Finally, call the corresponding accumulation function with `compounding: true`.

This is the most "correct" algorithm for a client to perform.

However, there are some simpler approaches:

1. Simply call the `harvest_yield` with `compounding: true`. If the call fails (due to running out of gas), then call `harvest_yield` with `snapshot_limit` set, and then try `compounding: true` again. Repeat until success.
2. Always perform one `snapshot_limit` call followed by one `compounding: true` call. (This is the approach I suggested above.)

PR #102, opened 2025-03-29
Implement storage management

### LINKED ISSUE #35: Storage management
# Justifying storage management in a multi-chain contract

Usually, NEAR smart contracts should implement some sort of storage management to ensure that the contract does not get soft-locked.

One argument against this for this particular contract (since the users are not necessarily supposed to hold any NEAR tokens) is to simply ensure that, upon contract deployment, there are minimum collateralize/supply amounts sufficient to cover the storage costs.

However, the argument against that is that it is a more complicated calculation to ensure the above, especially since borrow/collateral assets are not likely to be stable in price relative to NEAR, the currency of the storage cost. Therefore, the argument is that the gas sponsorship service should also be a storage sponsorship service.

# How should storage management be implemented?

A second issue: While it is clear who should pay for the storage consumed by individual positions (the borrower should pay for the storage consumed by his borrow position, and the supplier his supply position), it is less clear who should pay for the storage consumed by the snapshots, which are created as time elapses (by the account (un)lucky enough to be the first in the new time chunk) for the lifetime of the market.

There are a few ways to approach this:
1. **"As luck would have it" approach**: Simply have the account that chances to be the one creating the newest snapshot pay for it. It won't be especially expensive, though it is slightly gameable and not _completely_ fair, since every active position at any given point in time will _eventuallly_ use every snapshot as part of their calculations.
    1. **Modification**: Every time an account uses a snapshot as part of its calculation, transfer storage balance from that account to the account that created the snapshot, effectively _refunding_ them over time. _This is a high-complexity solution and I do not recommend it._
2. **"The past predicts the future" approach**: Keep track of the number of accounts that used the _last_ snapshot $n$ and charge each account that uses the contract in the _current_ snapshot for the amount of storage consumed by a single snapshot $\div n$. _Also decently complex._
3. **"Benevolent Donator For Life (BDFL)" approach**: Only have accounts pay for their own storage, and rely on the market curator to keep the contract from becoming softlocked. _This option is arguably a little bit less decentralized, though nothing prevents random users from funding the contract either. Regardless, the contract is definitely less self-sufficient w.r.t. long-term maintenance._
    1. **Modification**: Accounts pay for their own storage, plus e.g. 2 snapshots worth of storage when they sign up, offsetting the burden on the market curator.
4. **"Accuracy is the thief of joy" approach**: Charge the snapshot storage fee to user accounts _every time_ the account performs a stateful action. _This will almost definitely result in overcharging users._

peer2f00l:


petarvujovic98 (approved):
Seems reasonable!

PR #108, opened 2025-03-29
chore: Remove unnecessary promise results count check

petarvujovic98:
Removal of `promise_results_count` check because of this [test](https://github.com/near/near-sdk-rs/blob/09ab9ce5f9c0432d5b1914e0344f6d174956eb1a/near-sdk-macros/src/core_impl/code_generator/item_impl_info.rs#L266) and the corresponding [snapshot](https://github.com/near/near-sdk-rs/blob/09ab9ce5f9c0432d5b1914e0344f6d174956eb1a/near-sdk-macros/src/core_impl/code_generator/snapshots/private_method.snap#L9) which shows that there is a generated check for private functions which would make this invariant impossible to get to.

### DIFF
--- a/contract/market/src/impl_helper.rs
+++ b/contract/market/src/impl_helper.rs
@@ -45,14 +45,13 @@ impl Contract {
         account_id: AccountId,
         amount: BorrowAssetAmount,
     ) -> BorrowAssetAmount {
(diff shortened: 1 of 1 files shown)


peer2f00l:
There were a few other checks with equivalent guarantees, so I removed them as well.

peer2f00l (approved):

PR #109, opened 2025-03-29
feat: Simplify harvest yield args

petarvujovic98:
There was some logic which disallowed usage of both optional parameters, so I just created an enum which satisfies all three possible choices and lets the RPCs handle our first case (where we panicked - now the call wouldn't even get to the contract).

---

# API Changes

`harvest_yield` arguments are changed:

- `harvest_yield(compounding: true)` &rarr; `harvest_yield(mode: Compounding)`
- `harvest_yield(snapshot_limit: 123)` &rarr; `harvest_yield(mode: { SnapshotLimit: 123 })`

### DIFF
--- a/common/src/market/external.rs
+++ b/common/src/market/external.rs
@@ -89,7 +89,7 @@ pub trait MarketExternalInterface {
     /// harvested in previous, non-compounding `harvest_yield` calls) is
     /// deposited to the supply record, so it will contribute to future yield
     /// calculations.
-    fn harvest_yield(&mut self, mode: HarvestYieldMode) -> BorrowAssetAmount;
+    fn harvest_yield(&mut self, mode: Option<HarvestYieldMode>) -> BorrowAssetAmount;
 
     /// This value is an *expected average over time*.
     /// Supply positions actually earn all of their yield the instant it is
--- a/common/src/market/external.rs
+++ b/common/src/market/external.rs
@@ -1,4 +1,4 @@
-use near_sdk::{AccountId, Promise, PromiseOrValue};
+use near_sdk::{near, AccountId, Promise, PromiseOrValue};
 
 use crate::{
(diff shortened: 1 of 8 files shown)


common/src/market/external.rs
```diff
@@ -76,15 +85,11 @@ pub trait MarketExternalInterface {
...
+    /// If mode is set to `compounding`, the all of the yield (including any
     /// harvested in previous, non-compounding `harvest_yield` calls) is
     /// deposited to the supply record, so it will contribute to future yield
     /// calculations.
-    fn harvest_yield(
-        &mut self,
-        compounding: Option<bool>,
-        snapshot_limit: Option<u32>,
-    ) -> BorrowAssetAmount;
+    fn harvest_yield(&mut self, mode: HarvestYieldMode) -> BorrowAssetAmount;
```
peer2f00l: Should be an `Option` so that unparameterized calls still perform the default mode.

contract/market/src/impl_market_external.rs
```diff
@@ -312,25 +307,17 @@ impl MarketExternalInterface for Contract {
...
-                self.configuration
-                    .collateral_asset
-                    .transfer(predecessor.clone(), collateral_asset_amount),
-            )
-        };
+        let collateral_promise = (!collateral_asset_amount.is_zero()).then_some(
+            self.configuration
+                .collateral_asset
+                .transfer(predecessor.clone(), collateral_asset_amount),
+        );
```
peer2f00l: I think this is not a equivalent set of operations: I think that this will still create (and execute) the promise, it just will not create a data receipt dependency for callbacks (i.e. it will not act like a return value). So, it should probably be a `then` call instead of a `then_some` call. I will create a test to verify this.

peer2f00l (changes requested):

peer2f00l (approved):

PR #110, opened 2025-03-30
feat: Use `LookupMap` instead of `UnorderedMap` for storing positions

### LINKED ISSUE #90: Don't use iterable collections to store borrow and supply positions
This should save a lot on storage fees. However, it will also make it more difficult to see the list of borrow/supply positions without digging into either an indexer or into the raw storage keys.

petarvujovic98:
# API Changes
Removes the `list_borrows` and `list_supplys` methods

### DIFF
--- a/common/src/market/external.rs
+++ b/common/src/market/external.rs
@@ -33,9 +33,6 @@ pub trait MarketExternalInterface {
     fn list_snapshots(&self, offset: Option<u32>, count: Option<u32>) -> Vec<&Snapshot>;
     fn get_borrow_asset_metrics(&self) -> BorrowAssetMetrics;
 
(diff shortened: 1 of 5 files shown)


petarvujovic98:
@peer2f00l Do you want to have a service storing data to a DB for easier retrieval from the frontend? This would just be a convenience service, so doesn't affect the actual behavior?

peer2f00l:
@petarvujovic98 We do not, and you are correct that this does not affect the integral behavior of the contract.

I think we can set up such a service, or use a preexisting indexer service. I have actually been discussing this performance improvement with @SmolinPavel , and had not yet decided on whether to pull the trigger. Regardless, even an indexer service is not *necessarily* required, since the account IDs can be scraped from the storage keys on the contract.

peer2f00l (approved):

PR #112, opened 2025-03-31
Simplified snapshot handling

### LINKED ISSUE #97: Do not add current snapshot to snapshots list until finalization
Improve clarity about which snapshots can be used for interest/yield calculations.

This will also involve changing how timestamps on snapshots are used. Right now, the timestamp on a snapshot represents when the timestamp was created. After this issue is resolved, the timestamp on a snapshot should represent when the snapshot was _finalized_.

peer2f00l:
- Introduce `current_snapshot` field.
- Change semantics of timestamp field from "beginning of time chunk" to "end of time chunk."
- Associated algorithmic simplifications.

# API Changes

- New function: `get_current_snapshot()`
- Rename `get_snapshots_len` &rarr; `get_finalized_snapshots_len`
- Rename `list_snapshots` &rarr; `list_finalized_snapshots`

### DIFF
--- a/common/src/market/impl.rs
+++ b/common/src/market/impl.rs
@@ -1,7 +1,4 @@
-use near_sdk::{
-    collections::{LookupMap, UnorderedMap},
-    env, near, AccountId, BorshStorageKey, IntoStorageKey,
-};
+use near_sdk::{collections::LookupMap, env, near, AccountId, BorshStorageKey, IntoStorageKey};
 
 use crate::{
     asset::BorrowAssetAmount,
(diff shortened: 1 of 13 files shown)


common/src/market/impl.rs
```diff
@@ -88,48 +101,33 @@ impl Market {
...
-        };
-        self.snapshots.push(new_snapshot);
-        if let Some(previous_snapshot_index) = index.checked_sub(1) {
-            if let Some(previous_snapshot) = self.snapshots.get(previous_snapshot_index) {
-                MarketEvent::SnapshotFinalized {
-                    index: previous_snapshot_index,
-                    snapshot: previous_snapshot.clone(),
-                }
-                .emit();
+            std::mem::swap(&mut snapshot, &mut self.current_snapshot);
```
petarvujovic98, line 121: Nice one, I thought about using it but didn't get to refactor the code before you got to it!

common/src/snapshot.rs
```diff
@@ -6,7 +6,7 @@ use crate::{asset::BorrowAssetAmount, number::Decimal, time_chunk::TimeChunk};
 #[near(serializers = [borsh, json])]
 pub struct Snapshot {
     pub time_chunk: TimeChunk,
-    pub timestamp_ms: U64,
```
petarvujovic98, line 9: Definitely adds more clarity for the reader

petarvujovic98 (approved):
Thanks taking over here. I only did some basic work late last night (my time that is) but didn't get to it before you this morning.

peer2f00l:
@petarvujovic98 Thanks for getting it started!

PR #113, opened 2025-03-31
Cache snapshot interest rate

peer2f00l:
This should drastically improve the gas consumption of interest accumulation calculations for borrow positions.

# API Changes

- Removes `get_last_interest_rate`, since this field is now included in every snapshot. To get the same value, call `get_current_snapshot` and access the `interest_rate` field of the response.

### DIFF
--- a/common/src/borrow.rs
+++ b/common/src/borrow.rs
@@ -203,8 +203,43 @@ impl<M> BorrowPositionRef<M> {
 
 impl<M: Deref<Target = Market>> BorrowPositionRef<M> {
     pub fn with_pending_interest(&mut self) {
-        self.position.borrow_asset_fees.pending_estimate =
-            self.calculate_interest(u32::MAX).get_amount();
+        let mut pending_estimate = self.calculate_interest(u32::MAX).get_amount();
+        let prev_end_timestamp_ms = self.market.get_last_finalized_snapshot().end_timestamp_ms.0;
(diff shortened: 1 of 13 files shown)


petarvujovic98 (approved):
While this is better IMO, is there a case for still calculating this per request (via a non default API) in order to avoid cache corruption?

SmolinPavel (approved):

peer2f00l:
> While this is better IMO, is there a case for still calculating this per request (via a non default API) in order to avoid cache corruption?

@petarvujovic98 I'm not sure. If we did add that feature, that would incentivize users to choose the optimal mode (cached vs. non-cached) to perform their accumulations, which is not a good thing to incentivize, I think. Additionally, although it is not (currently) enforced with type design, the only place that snapshots are updated is within a single function, and that function ensures that the `interest_rate` field is properly updated.

Therefore, for at least _this_ version of the code, I think that cache corruption will not be an issue.

I should, however, add a task to more definitively enforce the updating of this field.

Edit: https://github.com/Templar-Protocol/contract-mvp/issues/114

peer2f00l:
As we can see from the gas report, this has handily evened-out the gas consumption metrics between borrow and supply accumulations. Supply took a bit of a hit from 24k&rarr;17k (since taking snapshots now guarantees an interest rate recalculation), while borrow nearly doubled from 8k&rarr;16k.

PR #145, opened 2025-05-07
fix: disallow collateralize and repay when eligible for liquidations

### LINKED ISSUE #135: Adding collateral should be disabled if the borrow is in liquidation [TD-5, TD-10, TD-11]
Avoids edge cases and race conditions during liquidation.

peer2f00l:


SmolinPavel (approved):

contract/market/src/impl_helper.rs
```diff
@@ -315,13 +328,45 @@ impl Contract {
         }
     }
 
+    pub const GAS_FT_ON_TRANSFER_COLLATERALIZE_01_CONSUME_PRICE: Gas = Gas::from_tgas(5);
```
peer2f00l, line 332: These need to be measured

PR #147, opened 2025-05-09
Introduce `Valuation` struct that tracks exponent separately from coefficient

### LINKED ISSUE #124: Track exponent separately from coefficient when computing valuations [TD-2]
This will help to avoid loss of precision when dealing with very small amounts/prices.

peer2f00l:


SmolinPavel (approved):

peer2f00l:
That's a hit on performance... 16k->12k :(

PR #155, opened 2025-06-09
Enforce minimum borrow amount during repayment

### LINKED ISSUE #129: Minimum borrow amount not enforced during repayment [TD-7]
A borrow position can be repaid down to just dust, which will not reset the expiration timer, etc. Minimum borrow amount should also be enforced here: if the amount repaid is not enough to fully repay the borrow, the minimum borrow amount should be retained and the rest refunded back to the user.

peer2f00l:


petarvujovic98 (commented):
Seems simple enough!

petarvujovic98 (approved):

petarvujovic98 (approved):

PR #158, opened 2025-06-10
chore: remove useless code

### LINKED ISSUE #138: Clarifications [TD-19]
1. Redundant code:

    https://github.com/Templar-Protocol/contract-mvp/blob/61bf6a2301c4eedb674828ee86e9ddb93ba4f55b/common/src/market/impl.rs#L273

2. ~~We should rename the "within collateral ratio" functions to "satisfies collateral ratio", because "within" makes it sound like the function should return `true` if the balance is _less than_ the ratio, which is not what it does.~~ **Already renamed.**

peer2f00l:


petarvujovic98 (approved):
Removing useless code is always nice

PR #159, opened 2025-06-12
Adds range limits for borrow, supply, withdrawal

### LINKED ISSUE #132: Add minimums for supply and withdrawal amounts [TD-12]
Prevents withdrawal queue DoS.

peer2f00l:
**This PR is as large as it is because enforcing these new limits requires adding a lot of additional locking logic.**

- Introduces the `[Valid]AmountRange` structs, which represent _inclusive_ ranges of token amounts.
- Changes the `*_maximum_*` and `*_minimum_*` configuration fields to use the aforementioned range struct, for a consistent API:
    - Introduces `borrow_range`, replacing `borrow_minimum_amount` and `borrow_maximum_amount`.
    - Introduces `supply_range`, replacing and augmenting `supply_maximum_amount`.
    - Introduces `supply_withdrawal_range`, introducing a configurable range of allowable amounts that can be withdrawn at a time.
 - Changes the deposit representation in supply records, unifying the active & inactive deposit fields under the same struct.
 - Introduces supply position amount locking, preventing supply positions from depositing above configured limits during failed withdrawals.
 - Renames `inactive` funds to `incoming` funds.
     - Incoming funds are segmented by snapshot index of activation instead of lumped into one.
 - Introduces `outgoing` lock amount (used during withdrawals).
 - Removes the `SupplyWithdrawalResolution` event type; the semantics of the `SupplyWithdrawn` are now consistent with the other events: it is only emitted on success.

### DIFF
--- a/common/src/supply.rs
+++ b/common/src/supply.rs
@@ -210,7 +210,17 @@ impl<'a> SupplyPositionGuard<'a> {
         Self(SupplyPositionRef::new(market, account_id, position))
     }
 
-    fn add_active(&mut self, amount: BorrowAssetAmount) {
+    fn activate_incoming(&mut self, amount: BorrowAssetAmount, activate_at_snapshot_index: u32) {
+        self.position
+            .borrow_asset_deposit
+            .incoming
+            .split(amount)
+            .unwrap_or_else(|| {
+                env::panic_str("Supply position `borrow_asset_deposit.incoming` underflow")
+            });
+        self.position
+            .borrow_asset_deposit
+            .activate_incoming_at_snapshot_index = activate_at_snapshot_index;
         self.position
             .borrow_asset_deposit
             .active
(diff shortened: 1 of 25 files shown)


petarvujovic98 (approved):

common/src/supply.rs
```diff
@@ -178,22 +210,75 @@ impl<'a> SupplyPositionGuard<'a> {
...
+            .borrow_asset_deposit
+            .incoming
+            .join(amount)
+            .unwrap_or_else(|| {
+                env::panic_str("Supply position `borrow_asset_deposit.incoming` overflow")
+            });
+        self.position
+            .borrow_asset_deposit
+            .activate_incoming_at_snapshot_index = activate_at_snapshot_index;
+    }
```
peer2f00l, line 292: Missing: should update market `borrow_asset_deposited_incoming`

SmolinPavel (approved):

petarvujovic98 (approved):
I like the bonus feature of sped up tests (at least somewhat)

PR #160, opened 2025-06-12
Allow anyone to run accumulations on positions

### LINKED ISSUE #125: Allow anyone to run accumulations on positions [TD-3]
This will allow:

- Liquidators to force fee accumulations.
- All positions to be up-to-date so that snapshots can be deleted.

peer2f00l:
In order to prevent abuse, we also need to account for fractional accumulation values, hence the introduction of the new field in the `Accumulator` and related structs.

### DIFF
--- a/common/src/accumulator.rs
+++ b/common/src/accumulator.rs
@@ -1,4 +1,4 @@
-use near_sdk::{near, require};
+use near_sdk::{json_types::U128, near, require};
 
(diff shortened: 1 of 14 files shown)


peer2f00l:
Not ready for review yet; still need to add tests.

petarvujovic98 (approved):

PR #161, opened 2025-06-17
Automatically delete empty positions

### LINKED ISSUE #118: Delete empty/zero positions so that accounts can recover storage deposit [TD-16]

peer2f00l:


SmolinPavel (approved):

common/src/market/impl.rs
```diff
@@ -188,6 +188,19 @@ impl Market {
...
+        if self
+            .supply_positions
+            .get(account_id)
+            .is_some_and(|p| p.can_be_removed())
+        {
+            self.supply_positions.remove(account_id);
+            true
+        } else {
+            false
+        }
```
petarvujovic98: Could a `and_then` work here instead of the if statement, it feels a bit ugly

petarvujovic98 (approved):
Overall looks clean, except for one small nit

PR #164, opened 2025-06-25
Position enumeration functions

### LINKED ISSUE #162: Enumeration of positions
Makes it easier to find liquidatable positions.

peer2f00l:


contract/market/src/impl_market_external.rs
```diff
@@ -179,7 +201,9 @@ impl MarketExternalInterface for Contract {
             "Withdrawal amount is outside of allowable range",
         );
 
-        self.withdrawal_queue.remove(&predecessor);
+        if self.withdrawal_queue.contains(&predecessor) {
+            self.withdrawal_queue.remove(&predecessor);
+        }
```
petarvujovic98: Why do we need to check before removing?

petarvujovic98 (approved):

contract/market/src/impl_market_external.rs
peer2f00l: Good catch, it's a leftover from when I was testing some gas stuff.

petarvujovic98 (approved):

peer2f00l:
Test failure is due to new scripts from registry PR not being compatible with the old format (which the testing contract for this PR was initialized with). So, I manually reset the testnet account and am re-running the failed step.

PR #166, opened 2025-06-27
chore: removes current snapshot from pending

### LINKED ISSUE #119: Do not include current snapshot yield in value returned by `get_supply_position`
Since it is quite unstable and not-yet-withdrawable, no point to include it, since it makes the frontend unable to accurately report withdrawable yield.

peer2f00l:


petarvujovic98 (approved):

peer2f00l:
Cleaned up manually

PR #170, opened 2025-07-06
feat: Add bots for liquidations and accumulation

petarvujovic98:


petarvujovic98:
@peer2f00l Currently the liquidator bot is missing the configuration for what a sufficient liquidation amount would be given a quote from a dex for the given asset pool.

The most recent commit adds a base implementation for Rhea finance as a swap backend but lacks path finding logic when a pool does not exist for an asset pair (meaning there should be multiple pools involved to get to the target asset).

An additional feature which will get delivered (but is a bit nastier of an implementation) is near intents. Their quote calculation happens on a separate RPC server (not a contract call) and the contract API is also a bit more complex to integrate into this automated flow.
I'll be updating the PR with those features and missing pieces, in the meantime feel free to leave comments here and to add/remove any features.

petarvujovic98:
@peer2f00l The code should be ready for review itself, but I have not been able to test it since the staging contract doesn't seem to be deploying properly (and from the action history seems like it's not unique to this PR) so if you have a manual deployment with the recent changes that would be helpful to test against as currently I am unable to read the testnet configurations since the `balance_oracle` was renamed to `price_oracle_configuration`.

joshuajbouw:
Going through this now, just be sure to drop the WIP if its ready to go.

Cargo.toml
```diff
@@ -1,15 +1,24 @@
...
 [workspace.dependencies]
-getrandom = { version = "0.2", features = ["custom"] }
+anyhow = "1.0.95"
+async-trait = "0.1.88"
+base64 = "0.22.1"
 borsh = { version = "1.5", features = ["unstable__schema"] }
+clap = { version = "4.5.23", features = ["derive", "env"] }
+getrandom = { version = "0.2", features = ["custom"] }
 hex = { version = "0.4.3", features = ["serde"] }
+futures = "0.3.31"
```
joshuajbouw: Place this before `getrandom` and after `clap`.

bots/src/accumulator.rs
```diff
@@ -0,0 +1,170 @@
...
+};
+
+#[derive(Debug, Clone, Parser)]
+pub struct Args {
+    /// Market to run accumulator for
+    #[arg(short, long, env = "MARKET_ACCOUNT_ID")]
+    pub markets: Vec<AccountId>,
+    /// Signer key to use for signing transactions
+    #[arg(short = 'k', long, env = "SIGNER_KEY")]
+    pub signer_key: SecretKey,
```
joshuajbouw, line 27: Would be good to get this from the NEAR files that are created with the `near` CLI if one isn't set.

bots/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+type BorrowPositions = HashMap<AccountId, BorrowPosition>;
+
+/// Helper constant for `TeraGas`.
+pub const TERA_GAS: u64 = 10u64.pow(12);
+/// Default gas for updating price data. 300 `TeraGas`.
+pub const DEFAULT_GAS: u64 = TERA_GAS * 300;
+/// One NEAR in yoctoNEAR.
+pub const ONE_NEAR: u128 = 10u128.pow(24);
+/// One yoctoNEAR in yoctoNEAR.
+pub const ONE_YOCTO_NEAR: u128 = 1u128;
```
joshuajbouw: These should be in the NEAR SDK if I'm not mistaken, except default gas.

bots/src/liquidator.rs
```diff
@@ -0,0 +1,418 @@
...
+        borrow: &AccountId,
+        liquidation_amount: U128,
+        nonce: u64,
+        block_hash: CryptoHash,
+    ) -> Transaction {
+        #[allow(clippy::unwrap_used, reason = "We know the serialization will succeed")]
+        let msg = serde_json::to_string(&DepositMsg::Liquidate(LiquidateMsg {
+            account_id: borrow.clone(),
+        }))
+        .unwrap();
```
joshuajbouw: Do avoid `unwrap()` in production code. Would be good to set a clippy lint to check for this.

bots/src/liquidator.rs
```diff
@@ -0,0 +1,418 @@
...
+        //  - Price impact
+        //  - Slippage
+        // All of this would be used in calculating both the optimal liquidation amount and wether to
+        // perform full or partial liquidation
+        let borrow_asset = &configuration.borrow_asset;
+        let collateral_asset = &configuration.collateral_asset;
+        let price_pair = configuration
+            .price_oracle_configuration
+            .create_price_pair(oracle_response)
+            .map_err(|e| anyhow::anyhow!("Failed to create price pair: {e}"))?;
```
joshuajbouw: `anyhow` should be avoided in production code. `thiserror` is more preferred, its more implicit. I'll post about it once here, as I see several instances.

bots/src/liquidator.rs
```diff
@@ -0,0 +1,418 @@
...
+            .await?;
+        let _quote_after_liquidate = self
+            .swap
+            .quote(
+                // TODO: Enable multitoken swaps
+                &collateral_asset.contract_id(),
+                &self.asset,
+                position.collateral_asset_deposit.into(),
+            )
+            .await?;
```
joshuajbouw: Just a question, what is the background to this? Could the return be helpful?

bots/src/liquidator.rs
```diff
@@ -0,0 +1,418 @@
...
+            json!({ "price_ids": price_ids, "age": age }),
+        )
+        .await
+    }
+
+    #[instrument(skip(self), level = "debug")]
+    async fn get_borrows(&self) -> anyhow::Result<BorrowPositions> {
+        let mut all_positions: BorrowPositions = HashMap::new();
+
+        let page_size = 100;
```
joshuajbouw: The page size might be too small. It feels like it may be a bit too many round trips, but may be a limitation of NEAR.

bots/src/liquidator.rs
```diff
@@ -0,0 +1,418 @@
...
+            all_positions.extend(page);
+            current_offset += page_size;
+            params["offset"] = current_offset.into();
+
+            if fetched < page_size {
+                break;
+            }
+        }
+
+        Ok(all_positions)
```
joshuajbouw: This should be a better implementation of the pagination.

Issues fixed:
- Doesn't handle a case when a page returns 0 results, potentially making an extra unnecessary request.
- Cloning and modifying JSON params object on each iteration, instead of just updating the offset value.
- Errors now propagate up the call stack instead of silently terminating the pagination loop.
- More accurate offset calculation. Now uses the actual number of items received rather than assuming we always get `page_size` items.

Potential improvement:
- All positions are held in memory, which could get problematic for large datasets. I'm not sure how large the datasets can get or are expected to get.

```suggestion
        let mut all_positions: BorrowPositions = HashMap::new();
        let page_size = 100;
        let mut current_offset = 0;

        loop {
            let params = json!({
                "offset": current_offset,
                "count": page_size,
            });

            let page = view::<BorrowPositions>(
                &self.client,
                self.market.clone(),
                "list_borrow_positions",
                params,
            ).await?;

            let fetched = page.len();

            if fetched == 0 {
                break;
            }
        
            all_positions.extend(page);
            current_offset += fetched;

            if fetched < page_size {
                break;
            }
        }

        Ok(all_positions)
```

bots/src/liquidator.rs
```diff
@@ -0,0 +1,418 @@
...
+                    "msg": msg,
+                })),
+                gas: DEFAULT_GAS,
+                deposit: ONE_YOCTO_NEAR,
+            }))],
+        })
+    }
+
+    #[instrument(skip(self), level = "debug")]
+    pub async fn try_liquidate(
```
joshuajbouw: Rename this to `liquidate` as `try_` isn't idiomatic Rust. It used to be popular in very early Rust, but these days its never used.

bots/src/liquidator.rs
```diff
@@ -0,0 +1,418 @@
...
+                }
+            })
+            .buffer_unordered(concurrency)
+            .collect::<Vec<_>>()
+            .await
+            .into_iter()
+            .collect::<anyhow::Result<Vec<_>>>()?;
+
+        Ok(())
+    }
```
joshuajbouw: This has a double collect which is unnecessary as its immediately discarded after accumulation.

```suggestion
    pub async fn run_liquidations(&self, concurrency: usize) -> anyhow::Result<()> {
        let configuration = self.get_configuration().await?;
        let oracle_response = self
            .get_oracle_prices(
                configuration.price_oracle_configuration.account_id.clone(),
                &[
                    configuration
                        .price_oracle_configuration
                        .borrow_asset_price_id,
                    configuration
                        .price_oracle_configuration
                        .collateral_asset_price_id,
                ],
                configuration.price_oracle_configuration.price_maximum_age_s,
            )
            .await?;

        futures::stream::iter(self.get_borrows().await?)
            .map(|(borrow, position)| {
                let oracle_response = oracle_response.clone();
                let configuration = configuration.clone();
                async move {
                    self.try_liquidate(borrow, position, oracle_response, configuration)
                        .await
                }
            })
            .buffer_unordered(concurrency)
            .try_for_each(|_result| async { Ok(()) })
            .await?;

        Ok(())
    }
```

bots/src/liquidator.rs
```diff
@@ -0,0 +1,418 @@
...
+                        .borrow_asset_price_id,
+                    configuration
+                        .price_oracle_configuration
+                        .collateral_asset_price_id,
+                ],
+                configuration.price_oracle_configuration.price_maximum_age_s,
+            )
+            .await?;
+
+        futures::stream::iter(self.get_borrows().await?)
```
joshuajbouw: You can add an early exit is `get_borrows` is `empty`.

bots/src/liquidator.rs
```diff
@@ -0,0 +1,418 @@
...
+            Liquidator::new(
+                client.clone(),
+                signer.clone(),
+                asset.clone(),
+                market.clone(),
+                swap.clone(),
+                args.timeout,
+            )
+        })
+        .collect::<Vec<_>>())
```
joshuajbouw: The explicit type annotation should be already implied to collect into a `Vec`. Simply `.collect()` should work, unless I am missing something.

joshuajbouw: This many clones makes me feel like that the implementation could certainly be made better. Perhaps as a new Struct which may include the client, signer, etc. per field, then a `Vec` of what needs to be liquidated. That would be far more efficient.

bots/src/near.rs
```diff
@@ -0,0 +1,139 @@
...
+pub async fn send_tx(
+    client: &JsonRpcClient,
+    signer: &InMemorySigner,
+    timeout: u64,
+    tx: Transaction,
+) -> anyhow::Result<FinalExecutionStatus> {
+    let (tx_hash, _size) = tx.get_hash_and_size();
+
+    let called_at = Instant::now();
+    let signature = signer.sign(tx.get_hash_and_size().0.as_ref());
```
joshuajbouw: This already is done above. Likewise, `get_hash_and_size` name implies that `get_hash` exists itself, so you can skip the return of `size`.

```suggestion
    let signature = signer.sign(tx_hash.as_ref());
```

bots/src/near.rs
```diff
@@ -0,0 +1,139 @@
...
+                            tx_hash,
+                        },
+                        wait_until: TxExecutionStatus::Final,
+                    })
+                    .await;
+                let elapsed = called_at.elapsed().as_secs();
+
+                if elapsed > timeout {
+                    bail!("Transaction timeout");
+                }
```
joshuajbouw: Given that there is a sleep of 2 seconds above, its very probable that the `elapsed` elapses over the timeout.

bots/src/near.rs
```diff
@@ -0,0 +1,139 @@
...
+
+                match status {
+                    Ok(status) => break status,
+                    Err(e) => match e.handler_error() {
+                        Some(RpcTransactionError::TimeoutError) => {}
+                        _ => Err(e)?,
+                    },
+                }
+            }
```
joshuajbouw: This should be better. It uses exponential backoff for polling (500ms -> 1s -> 2s -> 5s max) and timeout is checked *before* the sleep.

```suggestion
              if !matches!(e.handler_error(), Some(RpcTransactionError::TimeoutError)) {
                  return Err(e.into());
              }

              // Poll with exponential backoff
              let mut poll_interval = Duration::from_millis(500);
              const MAX_POLL_INTERVAL: Duration = Duration::from_secs(5);

              loop {
                  if Instant::now() >= deadline {
                      bail!("Transaction timeout after {}s", timeout);
                  }

                  tokio::time::sleep(poll_interval).await;

                  // Exponential backoff
                  poll_interval = std::cmp::min(poll_interval * 2, MAX_POLL_INTERVAL);

                  let status = client
                      .call(RpcTransactionStatusRequest {
                          transaction_info: TransactionInfo::TransactionId {
                              sender_account_id: signer.account_id.clone(),
                              tx_hash,
                          },
                          wait_until: TxExecutionStatus::Final,
                      })
                      .await;

                  match status {
                      Ok(status) => break status,
                      Err(e) => {
                          if !matches!(e.handler_error(), Some(RpcTransactionError::TimeoutError)) {
                              return Err(e.into());
                          }
                      }
                  }
              }
```

bots/src/swap.rs
```diff
@@ -0,0 +1,172 @@
...
+    RheaSwap,
+}
+
+impl SwapType {
+    #[must_use]
+    #[allow(
+        clippy::unwrap_used,
+        reason = "We know the contract IDs are valid NEAR account IDs."
+    )]
+    pub fn get_account_id(self, network: Network) -> AccountId {
```
joshuajbouw: Follow idiomatic Rust by naming this function `account_id`.

joshuajbouw (changes requested):
On first pass, I've highlighted a number of changes that can be made. Some idiomatic, some logic redos and others to enhance performance overall.

bots/src/liquidator.rs
petarvujovic98: > Doesn't handle a case when a page returns 0 results, potentially making an extra unnecessary request.

What is the consequence of not doing this explicitly since the `if fetched < page_size` would break anyway, trying to extend with an empty collection isn't an error.
> Cloning and modifying JSON params object on each iteration, instead of just updating the offset value.

I see the new approach creates a new object each time, which doesn't really save on memory anyway, but I do see it being more idiomatic
> More accurate offset calculation. Now uses the actual number of items received rather than assuming we always get page_size items.

While this would be more correct, it doesn't actually change the logic at all, so I don't see it as an issue, but can go with the new approach anyway

bots/src/liquidator.rs
petarvujovic98: Nice one, I did not know about the `try_for_each`

bots/src/liquidator.rs
petarvujovic98: I agree, this started off as a liquidator for a single market, and then due to my tight time budget only got squished into this form due to the need to support a collection of markets

bots/src/near.rs
petarvujovic98: Oversight on my part

bots/src/near.rs
petarvujovic98: Good catch

bots/src/accumulator.rs
petarvujovic98: We could, although since this is meant more as a demonstration of a liquidator bot, I think we can skip this for v0

bots/src/liquidator.rs
petarvujovic98: There is a lint, but disabled for this particular function call as we know that this will not panic - no user input, but I can swap it for a yield and move into thiserror

bots/src/liquidator.rs
petarvujovic98: It can, but I am not certain as to how liquidators usually assess profitability, this value would give them the ability to check if the liquidation would be profitable given the swap amount they would get after the liquidation

bots/src/liquidator.rs
petarvujovic98: Don't have time to handle this right now, but would be happy to do it in a follow up PR

petarvujovic98:
@joshuajbouw I've addressed most of the comments, so I'd appreciate another look at the PR if you are available

bots/src/accumulator.rs
joshuajbouw: Sure makes sense, will resolve for now.

bots/src/liquidator.rs
joshuajbouw: If it'll never fail, `expect` should be used then.

bots/src/accumulator.rs
```diff
@@ -0,0 +1,170 @@
...
+        futures::stream::iter(borrows)
+            .map(|(account_id, _)| async move { self.accumulate(account_id).await })
+            .buffer_unordered(concurrency)
+            .collect::<Vec<_>>()
+            .await
+            .into_iter()
+            .collect::<anyhow::Result<Vec<_>>>()?;
+
+        Ok(())
+    }
```
joshuajbouw: Noticed a double collect here, and added an early exit.

```suggestion
    #[instrument(skip(self), level = "info")]
    pub async fn run_accumulations(&self, concurrency: usize) -> anyhow::Result<()> {
        let borrows = self.get_borrows().await?;
        
        if borrows.is_empty() {
            return Ok(());
        }

        futures::stream::iter(borrows)
            .map(|(account_id, _)| async move { self.accumulate(account_id).await })
            .buffer_unordered(concurrency)
            .try_for_each(|_result| async { Ok(()) })
            .await?;

        Ok(())
    }
```

bots/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+                Network::Mainnet => "mainnet",
+                Network::Testnet => "testnet",
+            }
+        )
+    }
+}
+
+impl Network {
+    #[must_use]
+    pub fn get_rpc_url(self) -> &'static str {
```
joshuajbouw: Rename to `rpc_url`, and `&str` on return is fine, don't need to write `&'static`

joshuajbouw (changes requested):
Just a couple of things that I missed on last review, otherwise looks good.

joshuajbouw (approved):
LGTM!

PR #172, opened 2025-07-14
Use `U256` for underlying asset amounts

### LINKED ISSUE #51: Change underlying data type of assets to 256-bit integer instead of 128-bit
This will prepare the contract for dealing with assets on networks that use 256-bit integers to store balances (i.e. EVM).

peer2f00l:
I think this is a good idea, since we plan to add support for EVM assets, balances for which are 256 bits wide. However, as of yet, the NEAR Intents contract still uses 128 bit unsigned integers for balances. (I think. At least, that is [what the multi-token standard (which `intents.near` implements) specifies](https://nomicon.io/Standards/Tokens/MultiToken/Core#reference-level-explanation).)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,26 +183,14 @@ checksum = "b048fb63fd8b5923fc5aa7b340d8e156aec7ec02f0c78fa8a6ddc2613f6f71de"
 
 [[package]]
 name = "bitvec"
-version = "0.20.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7774144344a4faa177370406a7ff5f1da24303817368584c6206c8303eb07848"
-dependencies = [
- "funty 1.1.0",
- "radium 0.6.2",
- "tap",
- "wyz 0.2.0",
-]
-
-[[package]]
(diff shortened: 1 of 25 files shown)


joshuajbouw:
Might make sense to instead have an enum and using the appropriate value based on the protocol, then you have the best of both worlds. I am not sure of the drawbacks right now however, will need to dive into it.

peer2f00l:
For some reason, this is breaking tests in ways that I haven't had the time to dive in and understand yet. I am _fairly certain_ that this is a case of foreseeing a problem that may never actualize, since for most EVM-based assets decimals = 18, and u128 supports up to 38 decimal places, so it could only become a problem for memecoins with extremely high supply (or assets with decimals > 18). For now, it's worth noting, but I'm not sure if it needs to be a priority.

joshuajbouw:
If this is urgent, let me know and I'll prioritise a review. I need to consider probable issues and do a thorough and careful review of this. Aiming to get my task complete first regarding DefiLlama.

peer2f00l:
For now, this is not a priority to complete.

PR #173, opened 2025-07-17
Snapshot encapsulation

### LINKED ISSUE #114: Protect `interest_rate` on snapshots to ensure value is never stale

peer2f00l:


common/src/snapshot.rs
```diff
@@ -24,4 +57,16 @@ impl Snapshot {
...
+        self.interest_rate
+    }
+
+    pub const fn get_deposited_active(&self) -> BorrowAssetAmount {
+        self.deposited_active
+    }
+
+    pub const fn get_borrowed(&self) -> BorrowAssetAmount {
+        self.borrowed
+    }
```
joshuajbouw: Two things:
- `const` shouldn't be used here given the reference to `self`, it depends on runtime state unless this is configured as a constant.
- `get_` generally in Rust best practices `interest_rate` `deposited_active` and `borrowed` is what is going to be expected. I would go further to possibly even rename `borrowed` to something else that may not be confused with `borrow`, but should be ok. `get_` implies that information is being retrieved from a storage layer, an option, or a collection, which isn't the case here.

joshuajbouw (changes requested):
I will set this as a requested change to ensure that there is conformity to idiomatic Rust as per Rust style guide. Functionally, it is fine.

joshuajbouw (approved):
Looks good to me.

PR #175, opened 2025-07-17
chore: rename BalanceOracleConfiguration -> PriceOracleConfiguration

peer2f00l:
I don't know why it was named this, since it is not a balance oracle.

Note for @SmolinPavel , this will change the configuration.


joshuajbouw (approved):
Makes sense to me, `price_oracle_configuration` is preferable to me as well.

PR #176, opened 2025-07-17
fix: panic on borrow_asset_deposited_active overflow

peer2f00l:


peer2f00l:
Still need to write tests

common/src/market/impl.rs
```diff
@@ -143,7 +143,9 @@ impl Market {
                 .borrow_asset_deposited_incoming
                 .remove(&self.finalized_snapshots.len())
                 .unwrap_or(0.into());
-            self.borrow_asset_deposited_active.join(deposited_incoming);
+            self.borrow_asset_deposited_active
+                .join(deposited_incoming)
```
joshuajbouw: `checked_add` would be preferable, then you have an `Option<T>` with `None` on overflow. Its more explicit too.

joshuajbouw (approved):
Clean, I like it.

Issue #183, opened 2025-07-25
Liquid staking token price adapter


peer2f00l:
Native liquid staking tokens on NEAR (stNEAR, LiNEAR) provide their own redemption rates. Since this is an exact redemption rate (i.e. with infinite precision), it is preferable to use that and then just use the oracle price for NEAR to derive price data.

PR #188, opened 2025-07-30
LST Oracle Adapter Contract

peer2f00l:
#183

### DIFF
--- a/common/src/oracle/price_transformer.rs
+++ b/common/src/oracle/price_transformer.rs
@@ -7,7 +7,7 @@ use crate::number::Decimal;
 
 use super::pyth::{self, PriceIdentifier};
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 #[near(serializers = [json, borsh])]
 pub enum Action {
     NormalizeNativeLstPrice,
@@ -30,7 +30,7 @@ impl Action {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 #[near(serializers = [json, borsh])]
 pub struct Call {
     pub account_id: AccountId,
@@ -76,7 +76,7 @@ impl Call {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 #[near(serializers = [json, borsh])]
 pub struct PriceTransformer {
     pub price_id: PriceIdentifier,
--- a/common/src/oracle/price_transformer.rs
+++ b/common/src/oracle/price_transformer.rs
@@ -42,6 +42,8 @@ pub struct Call {
 }
 
 impl Call {
+    #[cfg(not(target_arch = "wasm32"))]
+    #[allow(clippy::unwrap_used)]
     pub fn new(
         account_id: &AccountIdRef,
         method_name: impl Into<String>,
@@ -56,6 +58,7 @@ impl Call {
         }
     }
 
+    #[cfg(not(target_arch = "wasm32"))]
     pub fn new_simple(account_id: &AccountIdRef, method_name: impl Into<String>) -> Self {
         Self::new(
             account_id,
@@ -90,10 +93,6 @@ impl PriceTransformer {
             call,
             action: Action::NormalizeNativeLstPrice,
         }
-    }
-
-    pub fn id(&self) -> PriceIdentifier {
-        PriceIdentifier(near_sdk::env::sha256_array(&borsh::to_vec(self).unwrap()))
     }
 }
 
--- a/common/src/oracle/price_transformer.rs
+++ b/common/src/oracle/price_transformer.rs
@@ -1,8 +1,6 @@
 use near_sdk::{
     json_types::{Base64VecU8, U64},
-    near,
-    serde::Serialize,
-    serde_json, AccountId, AccountIdRef, Gas, NearToken, Promise,
+    near, AccountId, Gas, NearToken, Promise,
 };
 
 use crate::number::Decimal;
@@ -45,25 +43,25 @@ impl Call {
     #[cfg(not(target_arch = "wasm32"))]
     #[allow(clippy::unwrap_used)]
     pub fn new(
-        account_id: &AccountIdRef,
+        account_id: &near_sdk::AccountIdRef,
         method_name: impl Into<String>,
-        args: impl Serialize,
+        args: impl near_sdk::serde::Serialize,
         gas: Gas,
     ) -> Self {
         Self {
             account_id: account_id.into(),
             method_name: method_name.into(),
-            args: serde_json::to_vec(&args).unwrap().into(),
+            args: near_sdk::serde_json::to_vec(&args).unwrap().into(),
             gas: gas.as_gas().into(),
         }
     }
 
     #[cfg(not(target_arch = "wasm32"))]
-    pub fn new_simple(account_id: &AccountIdRef, method_name: impl Into<String>) -> Self {
+    pub fn new_simple(account_id: &near_sdk::AccountIdRef, method_name: impl Into<String>) -> Self {
         Self::new(
             account_id,
             method_name,
-            serde_json::Value::Null,
+            near_sdk::serde_json::Value::Null,
             Gas::from_tgas(3),
         )
     }
--- a/common/src/oracle/price_transformer.rs
+++ b/common/src/oracle/price_transformer.rs
@@ -10,14 +10,14 @@ use super::pyth::{self, PriceIdentifier};
 #[derive(Clone, Debug, PartialEq, Eq)]
 #[near(serializers = [json, borsh])]
 pub enum Action {
-    NormalizeNativeLstPrice,
+    NormalizeNativeLstPrice { decimals: u32 },
 }
 
 impl Action {
     pub fn apply(&self, mut price: pyth::Price, input: Decimal) -> Option<pyth::Price> {
         match self {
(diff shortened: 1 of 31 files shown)


common/src/oracle/price_transformer.rs
```diff
@@ -0,0 +1,130 @@
...
+        match self {
+            Self::NormalizeNativeLstPrice => {
+                let scale_factor = input / Decimal::TEN_POW_24;
+
+                let price_is_negative = if price.price.0.is_negative() { -1 } else { 1 };
+                let abs_price_u128 = i128::from(price.price.0).unsigned_abs();
+                price.price.0 = price_is_negative
+                    * i64::try_from((abs_price_u128 * scale_factor).to_u128_floor()?).ok()?;
+                price.conf.0 = u64::try_from((price.conf.0 * scale_factor).to_u128_ceil()?).ok()?;
+                Some(price)
```
joshuajbouw, line 27: Two issues.

* Might be a hypothetical, but `.to_128_floor` may introduce truncated precision and `to_u128_ceil` may introduce bias.
* `.ok()` could add silent failing which could result in unintended failure on `None`.

contract/lst-oracle/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+        };
+
+        Owner::init(&mut self_, &env::predecessor_account_id());
+
+        self_
+    }
+
+    pub fn get_oracle_id(&self) -> AccountId {
+        self.oracle_id.clone()
+    }
```
joshuajbouw: Rename to `oracle_id` and its best to return always as a `&AccountId`, which later someone can choose to turn into an owned value later.

contract/lst-oracle/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+
+    // impl Pyth:
+
+    pub fn price_feed_exists(&self, price_identifier: PriceIdentifier) -> bool {
+        self.transformers.get(&price_identifier).is_some()
+    }
+
+    pub fn list_ema_prices_no_older_than(
+        &self,
+        price_ids: Vec<PriceIdentifier>,
```
joshuajbouw, line 108: Early exit on this `Vec` would be safe, unless its always expected to have `PriceIdentifier`s.

contract/lst-oracle/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+                near_sdk::log!("Transforming price: {price_id}");
+                let Some(entry) = self.transformers.get(&price_id) else {
+                    env::panic_str(&format!(
+                        "No transformer associated with price ID: {price_id}",
+                    ));
+                };
+                let Some(price) = oracle_result.get(&entry.price_id) else {
+                    env::panic_str(&format!(
+                        "Mapped price ID is not in oracle result: {price_id}",
+                    ));
```
joshuajbouw: Just a note on the logs and panic strings, in my experience strings take up a significant amount of data in a compiled contract to the point that we had a very large reduction in contract size. When we uniformed this at Aurora and truncated to only be IDs, this greatly reduced the footprint.

test-utils/src/controller/lst_oracle.rs
```diff
@@ -0,0 +1,78 @@
...
+        let contract = account.deploy(wasm).await.unwrap().unwrap();
+        contract
+            .call("new")
+            .args_json(json!({
+                "oracle_id": oracle_id,
+            }))
+            .transact()
+            .await
+            .unwrap()
+            .unwrap();
```
joshuajbouw, line 43: Transactions can fail for a number of reasons on deployment, its best to have it captured what failed.

test-utils/src/controller/lst_oracle.rs
```diff
@@ -0,0 +1,78 @@
...
+        &self.contract
+    }
+}
+
+impl LstOracleController {
+    pub async fn deploy(account: Account, oracle_id: &AccountId) -> Self {
+        static WASM: OnceCell<Vec<u8>> = OnceCell::const_new();
+
+        let wasm = WASM
+            .get_or_init(|| get_contract("templar_lst_oracle_contract", "contract/lst-oracle"))
```
joshuajbouw, line 31: From what I can tell, this expects a sync closure but get_contract is likely async, which could deadlock.

joshuajbouw (changes requested):
Just a couple of issues that I noted. Otherwise, looks good.

common/src/oracle/price_transformer.rs
peer2f00l: 1. I'm not sure we have too much of a choice here. I think the LST assets we're dealing with are _usually_ going to be used as collateral, so we should want to bias the price down (if at all), but that's not guaranteed.

    I'm thinking in the case of `price.price`, the direction of the bias is probably not as important as its consistency (i.e. I don't think it's a good idea to try to bias the transformation "cleverly," since this is just supposed to be an oracle).

    However, in the case of `price.conf`, I'm pretty sure that rounding up is the correct call, since this transformation cannot improve the resolution of the information it's transforming. (Note that `price.conf` is unsigned and represents ±error.)

    Regardless of which direction we round `price.price`, the rounding-up of `price.conf` should cover us.

1. The `.ok()` will fail only when `abs_price_u128 * scale_factor >= 2**128`. This is possible, but only in the case that `scale_factor > 2**64`, which seems exceedingly unrealistic for an LST, and I think it is OK for the calculation to fail in this case.

contract/lst-oracle/src/lib.rs
peer2f00l: Logs are there for debugging, so I will remove them before merging, thanks for pointing this out.

Would something like #62 be what you're referring to, otherwise? We haven't had to be too worried about binary size yet, though typed errors are always nicer.

test-utils/src/controller/lst_oracle.rs
peer2f00l: This is just in a test controller, so I'm OK with `.unwrap()`s and stuff like that. Happy to discuss if you disagree.

test-utils/src/controller/lst_oracle.rs
peer2f00l: This is `tokio::sync::OnceCell`, not `std::sync::OnceCell`.

peer2f00l:
Higher-level things to cover about this contract:

1. Currently it is single-owner (transferable), and only the owner can create price transformers.
2. Price transformers cannot be updated, removed, or overridden after they have been created.
3. The price IDs are arbitrarily chosen by the creator of the price transformer. (Alternatives could include a deterministic price ID function that hashes any combination of the transformer itself, the current account ID, the # of extant transformers on this contract, chain ID, etc.).

common/src/oracle/price_transformer.rs
joshuajbouw: 1. Had the same lines of thought running when going through it. I think it should be fine regardless, I don't necessarily see it as a red flag just a flag.
2. Sure in principle I do agree in a realistic setting its fine, but once fuzzy tests are deployed it would produce edge cases.

contract/lst-oracle/src/lib.rs
joshuajbouw: Yeah its that. When can get to it when we get to it, thats fine. We used stuff like "ERR_<someissue>" which was fairly general. Its possible to tack in what function it happened in as well by getting the name of the method which already is written as a string in the compiled code too for more clarity.

test-utils/src/controller/lst_oracle.rs
joshuajbouw: Oh - I missed it was a test, yeah this is fine.

peer2f00l:
Something else to do: forward more of the Pyth API functions in case we switch which one we're using later.

contract/lst-oracle/src/lib.rs
```diff
@@ -0,0 +1,189 @@
...
+                i += 1;
+
+                result.insert(
+                    price_id,
+                    price
+                        .clone()
+                        .and_then(|price| entry.action.apply(price, input)),
+                );
+            }
+        }
```
joshuajbouw, line 185: Changed to use built in `iter` and enumerate, `i + 1` because `i` starts at 0 with enumerate.

```suggestion
        for (i, price_id) in original_price_ids.iter().enumerate() {
            if let Some(price) = oracle_result.get(&price_id) {
                result.insert(price_id, price.clone());
            } else {
                let Some(entry) = self.transformers.get(&price_id) else {
                    env::panic_str(&format!(
                        "No transformer associated with price ID: {price_id}",
                    ));
                };
                let Some(price) = oracle_result.get(&entry.price_id) else {
                    env::panic_str(&format!(
                        "Mapped price ID is not in oracle result: {price_id}",
                    ));
                };
                let input = callback_result::<Decimal>(i + 1);

                result.insert(
                    price_id,
                    price
                        .clone()
                        .and_then(|price| entry.action.apply(price, input)),
                );
            }
        }
```

joshuajbouw (changes requested):
Small change and good to go.

contract/lst-oracle/src/lib.rs
peer2f00l: This is functionally different, since price IDs earlier in the list may not have instigated an additional promise, so it's not always its positional index + 1.

joshuajbouw: Ah - yeah, my mistake. Will change it now.

joshuajbouw: You are right - good point. Disregard this suggestion.

joshuajbouw (approved):
LGTM!

PR #189, opened 2025-07-31
chore: upgrade near-cli to v0.22.0

### LINKED ISSUE #187: Fix GH Actions by upgrading NEAR CLI
Example from failing workflow run: https://github.com/Templar-Protocol/contracts/actions/runs/16566250593/job/46852606446#step:11:55

Seems to be hitting this code in v0.19.0: https://github.com/near/near-cli-rs/blob/bfdf90c5784838db21552b15a5fd754031f02a42/src/common.rs#L1364

Latest version appears to handle without panic: https://github.com/near/near-cli-rs/blob/70b6d12bca235fde4a43ad2bcb8ae027e36db189/src/common.rs#L1421-L1440

peer2f00l:
It appears that there is a bug in nearcore that earlier versions of `near-cli-rs` were not able to handle, and it was breaking the GH Actions workflows. Fixed as of v0.22.0: https://github.com/near/near-cli-rs/pull/510

### DIFF
--- a/.github/workflows/deploy-production.yml
+++ b/.github/workflows/deploy-production.yml
@@ -21,7 +21,5 @@ jobs:
           cargo near deploy build-reproducible-wasm "${{ vars.NEAR_CONTRACT_PRODUCTION_ACCOUNT_ID }}" \
             without-init-call \
             network-config "${{ vars.NEAR_CONTRACT_PRODUCTION_NETWORK }}" \
-            sign-with-plaintext-private-key \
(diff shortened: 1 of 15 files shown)


petarvujovic98 (changes requested):
Looks like the `sign-with-plaintext-private-key` is not correctly set up

peer2f00l:
Seems like that was an issue with the GH actions cache, though there was another issue with `cargo near`, which should hopefully be fixed now.

petarvujovic98 (approved):

PR #191, opened 2025-08-04
Transaction Relayer (Gas Sponsor)

### LINKED ISSUE #13: Transaction gas sponsor
If the UX is going to be totally NEAR agnostic, the user-instigated transactions (e.g. from zero-balance implicit NEAR accounts) will need to be sponsored (meta transactions), so we will need a sponsorship service running.

Not an immediate priority.

peer2f00l:


contract/market/examples/gas_report.rs
petarvujovic98, line 1: Why is this test update part of the relayer PR?

service/relayer/migrations/20250815064726_init.up.sql
petarvujovic98, line 1: Just a nit, but usually keywords in SQL are capitalized, if you want to simplify the flow you could use a CLI I've built to format the SQL nicely - [sqlify](https://crates.io/crates/sqlify)

service/relayer/src/app/args.rs
```diff
@@ -0,0 +1,73 @@
...
+    #[arg(short = 'k', long, env = "SECRET_KEY", value_delimiter = ',')]
+    pub secret_key: Vec<SecretKey>,
+    /// Comma-separated list of allowed methods.
+    #[arg(long, env = "ALLOWED_METHODS", default_values_t = default_allowed_methods(), value_delimiter = ',')]
+    pub allowed_methods: Vec<String>,
+    /// Starting allowance in yoctoNEAR.
+    #[arg(long, env = "STARTING_ALLOWANCE_YOCTO", default_value = "0.25 NEAR")]
+    pub starting_allowance_yocto: NearToken,
+    /// Refresh the cached gas price after X seconds.
+    #[arg(long, env = "CACHE_GAS_PRICE_SECS", default_value = "600")]
```
petarvujovic98: I prefer using `default_value_t` for types that support it, since it gives you more compile time checks

service/relayer/src/app/mod.rs
```diff
@@ -0,0 +1,328 @@
...
+                    });
+                };
+
+                let Ok(msg) = serde_json::from_str::<DepositMsg>(transfer.args.msg()) else {
+                    return Err(PreconditionError::MsgDeserializationFailure { index });
+                };
+
+                if transfer.asset() == market_account_ids.borrow_asset {
+                    if !matches!(msg, DepositMsg::Supply | DepositMsg::Repay) {
+                        return Err(PreconditionError::InvalidMsgForAsset { index });
```
petarvujovic98: Do we want to add `expected` / `received` messages to these errors?

service/relayer/src/broom.rs
```diff
@@ -0,0 +1,67 @@
...
+                                continue;
+                            }
+                        };
+
+                        if let Err(e) = database.record_transaction(&account_id, &status).await {
+                            warn!("Broom error trying to automatically record transaction ({account_id}, {transaction_hash}): {e}");
+                        }
+                    }
+                } else {
+                    info!("Failed to fetch pending transactions.");
```
petarvujovic98: Shouldn't this be an `error` or `warn` log instead of `info`?

service/relayer/src/cache.rs
```diff
@@ -0,0 +1,171 @@
...
+                        #[allow(clippy::unwrap_used)]
+                        if let Ok(price) = fresh {
+                            sender.send(*price).unwrap();
+                        } else if let Some(price) = gas_price.stale() {
+                            tracing::warn!("Failed to fetch gas price, sending stale value.");
+                            sender.send(*price).unwrap();
+                        } else {
+                            // We should only ever *not* have a stale value on startup, so this should be a "fail-fast" case.
+                            tracing::error!("Failed to fetch gas price, no stale value cached.");
+                            panic!("Failed to fetch gas price.");
```
petarvujovic98: This will panic the spawned task, but we should also shut down the relayer itself if this happens, so I'd also pass in a `watch` channel for communicating with the main task that this failed, thus we can gracefully shut down anything else, or at the very least have better error messaging

service/relayer/src/client/database.rs
```diff
@@ -0,0 +1,392 @@
...
+        limit: i64,
+    ) -> Result<Vec<(AccountId, CryptoHash)>, sqlx::Error> {
+        let results = sqlx::query!(
+            "select
+                account_id,
+                pending_transaction_hash
+            from account
+            where pending_transaction_hash is not null
+            order by pending_transaction_issued_at asc
+            limit $1",
```
petarvujovic98: The tool I mentioned earlier could also be used in `vim` to pass it stdin and format strings too, although I usually have the `"` on a separate line from the rest of the SQL, I also have a snippet for highlighting SQL from these macros as SQL instead of rust strings if you want that feature

service/relayer/src/client/database.rs
```diff
@@ -0,0 +1,392 @@
...
+                }
+                .into())
+            } else {
+                Err(error::SetPendingTransactionError::UnknownError(
+                    account_id.to_owned(),
+                ))
+            }
+        } else {
+            Ok(())
+        }
```
petarvujovic98: Why not simply use a reversal of this branch to have the `Ok(())` value returned earlier and then have less indenting in the rest of the checks?

service/relayer/src/client/near.rs
```diff
@@ -0,0 +1,334 @@
...
+        let response = self
+            .client
+            .call(methods::query::RpcQueryRequest {
+                block_reference: Finality::Final.into(),
+                request: QueryRequest::ViewAccount { account_id },
+            })
+            .await?;
+
+        let QueryResponseKind::ViewAccount(account) = response.kind else {
+            unimplemented!("Invalid response kind");
```
petarvujovic98: Maybe `unreachable` is more suitable here?

service/relayer/src/main.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[cfg(not(unix))]
+    let terminate = std::future::pending::<()>();
+
+    tokio::select! {
+        () = ctrl_c => {},
+        () = terminate => {},
+    };
+
+    database.close().await;
+}
```
petarvujovic98: You could also listen to more signals like here: https://github.com/Sovereign-Labs/sovereign-sdk/blob/34a1c47432afc822127544aa0c818dab150360a7/crates/module-system/sov-modules-rollup-blueprint/src/native_only/mod.rs#L685

service/relayer/src/route/relay/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+#[allow(clippy::too_many_lines)]
+pub async fn relay(
+    State(app): State<App>,
+    Json(RelayRequest {
+        signed_delegate_action,
+        storage_deposit,
+        wait_until,
+    }): Json<RelayRequest>,
+) -> RelayResponse {
+    match app.check_and_calculate_gas(&signed_delegate_action).await {
```
petarvujovic98: Similar comment to what I left above, a nit about reducing indentation

petarvujovic98 (approved):
Overall looks solid, there are some niceties I'd change or make some code more readable, but those are not blockers from my POV

contract/market/examples/gas_report.rs
peer2f00l: Right... this is because an earlier iteration of the relayer included estimated gas values for each supported action. While I think this information is still good to have, it could be moved to a separate PR.

peer2f00l: Moved to https://github.com/Templar-Protocol/contracts/pull/215

peer2f00l:
Added basic SQL formatting; should add SQL format _checking_ in future PR. https://github.com/Templar-Protocol/contracts/issues/216

Issue #171, opened 2025-07-08
DefiLlama integration


royalf00l:
# DefiLlama Integration
 
## Description
Implement DefiLlama integration for the Templar smart contract protocol to enable TVL (Total Value Locked) and other metrics reporting, as outlined in the [DefiLlama project submission guide](https://docs.llama.fi/list-your-project/submit-a-project). Use an endpoint provided by pikespeak indexer for the adapter to fetch necessary data.

## Tasks
1. Create an adapter for Templar following DefiLlama’s [adapter requirements](https://docs.llama.fi/list-your-project/submit-a-project).
2. Integrate with the pikespeak indexer-provided endpoint to retrieve protocol data (e.g., TVL, balances).
3. Ensure the adapter exposes data in the format expected by DefiLlama (JSON structure with TVL by chain and other metrics).
4. Test the adapter locally to verify data accuracy and endpoint compatibility.
5. Submit the adapter to DefiLlama for review and inclusion.
6. Document the integration process and endpoint usage in the project’s README.

## Requirements
- TVL (total deposits (collateral assets + borrow assets) - outstanding loans) calculated for each market and each chain as well as aggregate
- Protocol Description and links to website, github and twitter
- All metrics tracked in [this example](https://defillama.com/protocol/aave?borrowed_tvl=true&devsMetrics=true&devsCommits=true&revenue=true&fees=true&contributersMetrics=true&contributersCommits=true). In addition to TVL, we want to track Borrowed, Revenue, Fees, all developer metrics.
- Ensure the adapter is lightweight and optimized for indexer data retrieval.
- Handle errors gracefully (e.g., indexer downtime, invalid data).
- Follow DefiLlama’s security and data accuracy guidelines.

## Acceptance Criteria
- Adapter successfully fetches and formats data from the indexer endpoint.
- Data is correctly reported to DefiLlama (verified via test submission).
- Integration is documented, and tests pass without errors.

## References
- [DefiLlama Submission Docs](https://docs.llama.fi/list-your-project/submit-a-project)
- Indexer endpoint: [https://pikespeak.ai/api/v1/templar/protocol-metrics](https://pikespeak.ai/api/v1/templar/protocol-metrics)

royalf00l:
[templar-api.pdf](https://github.com/user-attachments/files/21166167/templar-api.pdf)

Pikespeak Indexer API Endpoints: 
https://pikespeak.ai/api/v1/templar/protocol-metrics
[...]

PR #192, opened 2025-08-04
feat: collateral tracking

joshuajbouw:
In order to support #171 (DefiLlama integration) we need to be able to expose the metrics of collateral standalone, as well as in the snapshot as well. This would then allow external parties to be able to calculate using the correct TVL method. TVL (total deposits (collateral assets + borrow assets) - outstanding loans)

### DIFF
--- a/contract/market/tests/snapshot.rs
+++ b/contract/market/tests/snapshot.rs
@@ -1,10 +1,6 @@
-use std::{collections::VecDeque, time::Duration};
+use std::time::Duration;
 use templar_common::{
-    asset::{BorrowAssetAmount, CollateralAssetAmount},
-    dec,
-    fee::Fee,
-    interest_rate_strategy::InterestRateStrategy,
-    time_chunk::TimeChunkConfiguration,
+    dec, fee::Fee, interest_rate_strategy::InterestRateStrategy, time_chunk::TimeChunkConfiguration,
 };
 use test_utils::*;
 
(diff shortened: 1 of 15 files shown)


joshuajbouw:
@peer2f00l few things that I am unsure about:
- Where is the best appropriate place to add to the tracking of collateral deposits in `Contract` and `Snapshot`?
- Does `inflight` matter like it does in the context of borrowed?

peer2f00l:
Add collateral step is here:

https://github.com/Templar-Protocol/contracts/blob/b167daf5844e1196ef9899d4147eaf7be4a26ed5/contract/market/src/impl_helper.rs#L40-L70

Withdraw collateral step is here:

https://github.com/Templar-Protocol/contracts/blob/b167daf5844e1196ef9899d4147eaf7be4a26ed5/contract/market/src/impl_market_external.rs#L119-L154

You should be able to perform just about all of the necessary calculations as-we-go by adding logic in those two places. (And to the `snapshot()` function, I suppose.)

peer2f00l (changes requested):
Please write tests to ensure the correctness of the new field as it appears in snapshots.

joshuajbouw:
Added tests and encapsulation on snapshot, which is better to use defined functions so that future updaters don't accidentally make mistakes.

contract/market/tests/snapshot.rs
```diff
@@ -0,0 +1,544 @@
...
+
+    // Interest rate should reflect utilization
+    assert!(
+        !last.interest_rate().is_zero(),
+        "Interest rate should be positive with borrowing activity",
+    );
+}
+
+#[tokio::test]
+async fn snapshot_at_time_boundaries() {
```
peer2f00l, line 420: This test doesn't actually do stuff at the snapshot boundaries since most of the contract calls take multiple callbacks and the block time for `near-workspaces` tests is 1 second.

peer2f00l (commented):
Tweaked some of the tests. In general, I think they should be more precise (checking actual values instead of just "is non-zero after deposit" or "is different from the previous state after X action").

joshuajbouw:
> Tweaked some of the tests. In general, I think they should be more precise (checking actual values instead of just "is non-zero after deposit" or "is different from the previous state after X action").

Do agree, they should've been more tightly bound. It was mostly the interest that was (understandably) inconsistent, but it could yield false positives.

contract/market/tests/snapshot.rs
joshuajbouw: Ok, made the window larger, and tightened it.

joshuajbouw:
Added priority high since this is blocking another priority-high #171.

peer2f00l (approved):

PR #200, opened 2025-08-10
feat: Fetch all deployments from registry

### LINKED ISSUE #199: Fetch all deployments from registry list in liquidator bot

petarvujovic98:
This PR implements the logic to fetch from registries in order to get a list of deployments from them.
In addition to this it also improves swap logic by checking if the liquidator account already has enough funds for that asset and adds a new `should_liquidate` method to be implemented by each liquidator as an additional heuristic.

### DIFF
--- a/bots/README.md
+++ b/bots/README.md
@@ -26,7 +26,7 @@ Running the Bot:
 
 ```bash
 liquidator-service \
-    --markets market1.testnet \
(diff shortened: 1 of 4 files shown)


joshuajbouw (approved):
Looks good to me!

peer2f00l (approved):
LGTM

Issue #204, opened 2025-08-14
Add multi token support to the liquidator bot


joshuajbouw:
Currently the liquidator only supports NEP-141 fungible tokens via a single AccountId. This limits the bot to liquidating positions with standard NEAR fungible tokens.

**Goal**: Extend support to NEP-245 multitoken assets, which are identified by both a contract_id and token_id. This will enable the liquidator to work with multitoken contracts and more complex asset types.

**Changes needed:**

- Replace single asset: AccountId in CLI args with FungibleAssetKind enum supporting both NEP-141 and NEP-245 formats
- Update balance checking, transfer calls, and swap logic to handle both asset types
- Support CLI format: "account.near" (NEP-141) or "contract.near:token_id" (NEP-245)

This change enables the liquidator to work with a broader range of NEAR ecosystem assets and prepares the foundation for more sophisticated liquidation strategies.

peer2f00l:
Completed in #220

PR #205, opened 2025-08-14
feat: add multi token support to bot

joshuajbouw:
Currently the liquidator only supports NEP-141 fungible tokens via a single AccountId. This limits the bot to liquidating positions with standard NEAR fungible tokens.

Will follow up with tests, once we have essential parts done with the bot as theres numerous working pieces right now.

Related to #204.


bots/src/liquidator.rs
```diff
@@ -169,19 +168,40 @@ impl<S: Swap> Liquidator<S> {
...
+                "mt_transfer_call".to_string(), // <-- Different method
+                json!({
+                    "receiver_id": self.market,
+                    "token_id": format!("nep141:{}", token_id),
+                    "amount": liquidation_amount,
+                    "msg": msg,
+                }),
+            ),
+        };
+
```
peer2f00l: This is probably cleaner if added as a method to `FungibleAsset`.

bots/src/types.rs
```diff
@@ -0,0 +1,61 @@
+use near_account_id::AccountId;
+use std::str::FromStr;
+
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum FungibleAssetKind {
```
peer2f00l: Prefer to use `FungibleAsset` from `common` crate.

`common::asset::FungibleAssetKind` is private intentionally, because it is not intended to be used directly. `FungibleAsset` should be preferred because it guarantees that amounts of one asset will never be mistaken one for another.

bots/src/swap.rs
```diff
@@ -109,61 +122,86 @@ enum SwapRequestMsg {
...
 
+        // Need to determine which contract to call for the transfer
+        // For swaps, we typically transfer the input token to the swap contract
+        let (receiver_id, method_name, transfer_args) = match from {
+            FungibleAssetKind::Nep141(account_id) => (
+                account_id.clone(),
+                "ft_transfer_call".to_string(),
+                json!({
+                    "receiver_id": self.contract,
+                    "amount": amount,
```
peer2f00l: In this code (and the analogous multitoken case below) the `amount` is the amount of the `from` token being sent. However, the `SwapRequestMsg` treats `amount` as an `output_amount`, indicating that it should be an amount of the `to` token. Which is it?

peer2f00l (changes requested):

petarvujovic98 (commented):
Generally looks good (besides the comments already made by @peer2f00l)

bots/src/liquidator.rs
```diff
@@ -211,18 +231,42 @@ impl<S: Swap> Liquidator<S> {
...
+                FungibleAssetKind::from(account_id)
+            } else if let Some((contract_id, token_id)) =
+                configuration.collateral_asset.clone().into_nep245()
+            {
+                FungibleAssetKind::from((contract_id, token_id))
+            } else {
+                return Err(LiquidatorError::StandardSupportError(
+                    "Unsupported collateral asset type".to_string(),
+                ));
+            };
```
petarvujovic98: I'd also make methods around this instead of repeating

bots/src/liquidator.rs
```diff
@@ -359,14 +403,33 @@ impl<S: Swap> Liquidator<S> {
...
+                        "account_id": self.signer.account_id,
+                        "token_id": format!("nep141:{}", token_id)
+                    }),
+                )
+                .await
+            }
+        }
         .map_err(LiquidatorError::FetchBalanceError)?;
         Ok(balance)
     }
```
petarvujovic98: This should likely also become a method on the `FungibleAssetKind` type instead

bots/src/types.rs
joshuajbouw: Using the common crate FungibleAsset<T>, thats fine, I just was hesitant because the common crate is more closely associated with contract methods themselves, and could lead to someone mistakenly using the transfer function as part of FungibleAsset<T> to transfer the value, when we need to instead be using the RPC.

Making an RPC common crate in the future, or simply just feature gating the common crate so that the same methods are used for RPC, and not contract, could be a possibility.

joshuajbouw:
Going to convert this to draft for now. There is a bit more issues that I am seeing that I need clarification on which I had notified Peter about.

bots/src/liquidator.rs
```diff
@@ -211,18 +209,15 @@ impl<S: Swap> Liquidator<S> {
...
-        let collateral_asset = configuration.collateral_asset.contract_id();
+        let collateral_asset = configuration.collateral_asset;
 
         let liquidation_amount = self
             .liquidation_amount(&position, &oracle_response, configuration)
             .await?;
 
-        let swap_output_amount = if self.asset.as_ref() == &borrow_asset {
-            let asset_balance = self.get_asset_balance(self.asset.as_ref().clone()).await?;
+        // This implies that the asset set by user can either be borrow or collateral then?
```
petarvujovic98: Can be any asset, isn't restricted to the assets of the market

bots/src/swap.rs
```diff
@@ -8,23 +13,24 @@ use near_primitives::{
...
+use near_sdk::{json_types::U128, near, serde_json, serde_json::json, AccountId, NearToken};
+use templar_common::asset::{BorrowAsset, CollateralAsset, FungibleAsset};
 
 #[async_trait::async_trait]
 pub trait Swap {
     /// Quotes the amount of `from` token to `to` token.
-    async fn quote(&self, from: &AccountId, to: &AccountId, amount: U128) -> RpcResult<U128>;
+    async fn quote(
+        &self,
+        from: &FungibleAsset<CollateralAsset>,
```
petarvujovic98: Not really a `CollateralAsset` as it can be any asset the user provides

joshuajbouw: Yeah this is being changed as we speak, don't worry. A From + To would work better in this case, then we can support the existing logic and prevent accidental arith on the wrong assets.

bots/src/liquidator.rs
joshuajbouw: Given that now we are using the common crate, I want to prevent adding RPC logic in that crate.

bots/src/liquidator.rs
```diff
@@ -211,34 +210,37 @@ impl<S: Swap> Liquidator<S> {
...
+        let collateral_asset_id = configuration.collateral_asset.as_asset_id();
 
         let liquidation_amount = self
-            .liquidation_amount(&position, &oracle_response, configuration)
+            .liquidation_amount(&position, &oracle_response, &configuration)
             .await?;
 
-        let swap_output_amount = if self.asset.as_ref() == &borrow_asset {
-            let asset_balance = self.get_asset_balance(self.asset.as_ref().clone()).await?;
+        let swap_output_amount = if from_asset_id == borrow_asset_id {
```
peer2f00l: Maybe we should change the equality semantics of `FungibleAsset<T>` to allow for different asset classes to be compared to one another. (Like a `FungibleAsset<A> == FungibleAsset<B>` comparison.)

Would that eliminate the need to introduce yet another `AssetId` type?

common/src/asset.rs
```diff
@@ -62,6 +71,38 @@ impl<T: AssetClass> FungibleAsset<T> {
         }
     }
 
+    pub fn transfer_call_params(
+        &self,
+        receiver_id: &AccountId,
+        amount: U128,
+        msg: &str,
+    ) -> TransferCallParams {
```
peer2f00l: This function could just return a `FunctionCallAction` instead of introducing a new type.

common/src/asset.rs
```diff
@@ -159,10 +210,26 @@ impl<T: AssetClass> Display for FungibleAsset<T> {
...
+        }
+        Ok(FungibleAsset::nep141(AccountId::try_from(s.to_string())?))
+    }
+}
+
 mod sealed {
     pub trait Sealed {}
 }
-pub trait AssetClass: sealed::Sealed + Copy + Clone {}
+pub trait AssetClass: sealed::Sealed + Copy + Clone + Send {}
```
peer2f00l: Might be best to remove the sealing on this trait. If we're only dealing with borrow & collateral I think it makes more sense, but I think there's little argument to be had for this if we're introducing new asset classes.

common/src/asset.rs
```diff
@@ -159,10 +210,26 @@ impl<T: AssetClass> Display for FungibleAsset<T> {
     }
 }
 
+impl<T: AssetClass> FromStr for FungibleAsset<T> {
+    type Err = near_account_id::ParseAccountError;
```
peer2f00l:
```suggestion
    type Err = <AccountId as FromStr>::Err;
```

Eliminates the need to introduce the `near-account-id` dependency.

common/src/asset.rs
```diff
@@ -159,10 +210,26 @@ impl<T: AssetClass> Display for FungibleAsset<T> {
...
+        if let Some((contract_id, token_id)) = s.split_once(':') {
+            if let Some(token_id) = token_id.strip_prefix("nep245:") {
+                return Ok(FungibleAsset::nep245(
+                    AccountId::try_from(contract_id.to_string())?,
+                    token_id.to_string(),
+                ));
+            }
+        }
+        Ok(FungibleAsset::nep141(AccountId::try_from(s.to_string())?))
+    }
```
peer2f00l: This is not able to parse the strings produced by the `Display` implementation. That's probably relevant parity.

common/src/asset.rs
```diff
@@ -1,15 +1,24 @@
...
     json_types::U128,
     near,
     serde_json::{self, json},
     AccountId, Gas, NearToken, Promise,
 };
+use std::str::FromStr;
+use std::{fmt::Display, marker::PhantomData};
 
-use crate::number::Decimal;
+pub type AssetId<'a> = (&'a AccountId, Option<&'a String>);
```
peer2f00l:
```suggestion
pub type AssetId<'a> = (&'a near_sdk::AccountIdRef, Option<&'a str>);
```

bots/src/swap.rs
```diff
@@ -8,23 +12,24 @@ use near_primitives::{
...
+    ) -> RpcResult<U128>;
 
     /// Swaps `from` token to `to` token.
-    async fn swap(
+    async fn swap<A: AssetClass, B: AssetClass>(
         &self,
-        from: &AccountId,
-        to: &AccountId,
+        from: FungibleAsset<A>,
+        to: FungibleAsset<B>,
```
peer2f00l: Should these be generic, or should they use the `FromAsset` and `ToAsset` asset classes that were also introduced?

joshuajbouw: I could remove the generic of it now since the latest comment does coerce, so I'll do that now.

common/src/asset.rs
joshuajbouw: It already is a dependency, but would remove the dependency for external crates, so yes good chance.

bots/src/liquidator.rs
joshuajbouw: I had considered that, and was battling in my mind if its a good idea to compare other asset classes with each other. That led me to compare standard types, just in case. But ultimately, we are just comparing account IDs, which should be fine.

I'll double check if it does remove the need for the AssetId, if so, then I'll add this capability.

joshuajbouw: The only problem that I see is that we do need to check if something is a NEP-141 or NEP-245 indiscriminately without passing an account ID and checking if it matches. I'll keep `AssetId` for now just for the `get_asset_balance` method, even though I don't like it too much, but it unblocks me.

common/src/asset.rs
joshuajbouw: We could, so others can implement it externally for their own purposes, will keep as is for now however.

joshuajbouw:
I had to move the `create_function_call_action` to bots/src/near.rs because there is a compilation issue when it tries to compile near_primitives crate in the common crate. I really want to get this over the line, so this can be addressed at a later stage if needed.

Issue #206, opened 2025-08-14
Only run contract Github flows if contracts are changed


joshuajbouw:
Its a bit excessive when we keep on needing to run the contract flows when working on code unrelated to the contract like the bots. I propose a change to improve this by adjusting the Github workflows relating to deployments to only run when any of those files had changed in the PR.

**Goal:** Adjust Github actions gas-report, deploy-staging, undeploy-staging to only run if contract had been changed.

**Changes Needed:**
* Update `gas-report.yml` Github Action to watch if changes are made in `common` and `contract` folders, then run the flow.
* Update `deploy-staging.yml` Github Action to watch if changes are made in `common` and `contract` folders, then run the flow.
* Update `undeploy-staging.yml` Github Action to watch if changes are made in `common` and `contract` folders, then run the flow.

PR #207, opened 2025-08-14
ci: improve workflows to only run on PR when contract logic is changed

joshuajbouw:
Its a bit excessive when we keep on needing to run the contract flows when working on code unrelated to the contract like the bots. I propose a change to improve this by adjusting the Github workflows relating to deployments to only run when any of those files had changed in the PR.

This fixes those issues by checking contract changes, or previous stages if they had run.

Related: #206

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -23,8 +23,9 @@ jobs:
             contracts:
               - 'common/**'
               - 'contract/**'
+              - 'mock/**'
+              - 'script/**'
               - '**/Cargo.toml'
-              - '**/Cargo.lock'
 
   test:
     uses: ./.github/workflows/test.yml
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -25,7 +25,6 @@ jobs:
               - 'contract/**'
               - 'mock/**'
               - 'script/**'
-              - '**/Cargo.toml'
 
   test:
     uses: ./.github/workflows/test.yml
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-staging.yml
```diff
@@ -3,17 +3,44 @@ on:
...
+      - name: Check for changes
+        uses: dorny/paths-filter@v3
+        id: changes
+        with:
+          filters: |
+            contracts:
+              - 'common/**'
+              - 'contract/**'
+              - '**/Cargo.toml'
+              - '**/Cargo.lock'
```
peer2f00l: Should probably include `mock` and `script` to catch test & CI related changes as well.

peer2f00l (changes requested):
One comment, otherwise seems good!

.github/workflows/deploy-staging.yml
joshuajbouw: Fixed in https://github.com/Templar-Protocol/contracts/pull/207/commits/c65f6481521e2d6f2e8527d4888273c25d229ea2

joshuajbouw:
@peer2f00l Added your request, makes sense. I also had removed Cargo.lock because Cargo.lock changes -could- happen on a NEAR smart contract which does change the gas costs and could cause tests to fail, and change gas costs. I should also remove Cargo.toml.

There is a caveat, that adding a dependency to say, bots would trigger a deployment but unsure how to get around that, unless the bots were in a different repo.

peer2f00l (approved):

Issue #209, opened 2025-08-18
Add automated daily snapshot collection for deployed market contracts


joshuajbouw:
Currently there's no automated way to trigger `get_current_snapshot` calls across all deployed market contracts. This limits our ability to maintain up-to-date analytics data and requires manual intervention to collect market snapshots.

**Goal**: Implement an automated daily snapshot collection system that discovers all deployed contracts from the registry and calls `get_current_snapshot` on each one via blockchain transactions.

**Changes needed**:

- Create bash script using NEAR CLI that fetches all deployed contracts from `v1.tmplr.near` registry
- Support pagination for contract discovery
- Make transaction calls to `get_current_snapshot` on each contract to write analytics data to chain
- Support CLI format: --account account.near --private-key "ed25519:..."

This change enables automated analytics data collection and ensures our monitoring systems have consistent, up-to-date market snapshots without manual intervention.

Follow up, this needs to be added as a cron job.

PR #210, opened 2025-08-18
feat: add snapshot script for market contracts

joshuajbouw:
Currently there's no automated way to trigger `get_current_snapshot` calls across all deployed market contracts. This limits our ability to maintain up-to-date analytics data and requires manual intervention to collect market snapshots.

This works adds in the ability to enable us to setup a cron job.

Note, that I hadn't tested this out yet myself but it should work from what I can tell. If a reviewer can test it once, that would be amazing as I work to set this up as a cron job. I had also tried to retain consistency with the other scripts, making use of the utils.

Related: #209

### DIFF
--- a/script/ci/record_snapshot.sh
+++ b/script/ci/record_snapshot.sh
@@ -1,0 +1,199 @@
+#!/usr/bin/env bash
+set -e
+
+SCRIPT_DIR=$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")
+source "$SCRIPT_DIR/utils.sh"
+
+parse_args "--network:NETWORK,--registry:REGISTRY_ID,--log-file:LOG_FILE,--delay:DELAY,--account:ACCOUNT_ID,--private-key:PRIVATE_KEY,--page-size:PAGE_SIZE" "$@"
+
+if [ -z "$NETWORK" ]; then
+    NETWORK="testnet"
+fi
+
+if [ -z "$REGISTRY_ID" ]; then
+    REGISTRY_ID="v1.tmplr.near"
+fi
+
+if [ -z "$LOG_FILE" ]; then
+    LOG_FILE="/tmp/near-snapshots-$(date +%Y-%m-%d).log"
+fi
+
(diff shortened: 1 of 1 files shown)


script/ci/record_snapshot.sh
peer2f00l, line 1: Other scripts use `-` as a word separator in the file name.

script/ci/record_snapshot.sh
```diff
@@ -0,0 +1,199 @@
...
+    local all_contracts=""
+    local from_index=0
+    local total_fetched=0
+
+    while true; do
+        log "Fetching batch starting from index $from_index (page size: $PAGE_SIZE)..."
+
+        local contracts_json
+        contracts_json=$(near contract call-function as-read-only "$REGISTRY_ID" \
+            get_all_deployments \
```
peer2f00l, line 60: This function does not exist.

script/ci/record_snapshot.sh
```diff
@@ -0,0 +1,199 @@
...
+            json-args "{\"from_index\": $from_index, \"limit\": $PAGE_SIZE}" \
+            network-config "$NETWORK" 2>/dev/null)
```
peer2f00l, line 62: This command hangs because it is waiting for the block identifier to view; ditto for other commands.

peer2f00l (changes requested):

script/ci/record_snapshot.sh
```diff
@@ -0,0 +1,199 @@
...
+if [ -z "$NETWORK" ]; then
+    NETWORK="testnet"
+fi
+
+if [ -z "$REGISTRY_ID" ]; then
+    REGISTRY_ID="v1.tmplr.near"
+fi
+
+if [ -z "$LOG_FILE" ]; then
+    LOG_FILE="/tmp/near-snapshots-$(date +%Y-%m-%d).log"
```
peer2f00l, line 18: Might want to include the time here as well. Currently duplicate runs will simply append to the previous run's log file.

PR #219, opened 2025-09-06
Partial liquidations

### LINKED ISSUE #213: Partial liquidations
Allow a liquidator to liquidate any fraction of a position. This should be applied proportionally, and give the liquidator the maximum discount available.

For example, if the liquidatable collateral is worth $100, and the maximum spread is 10%, if a liquidator sends the contract $45, they should receive $50 worth of collateral from the contract. (assuming my math is correct)

peer2f00l:


peer2f00l:
~~Need to confirm that partial liquidations up to maintenance MCR is implemented correctly...~~

**EDIT**

Confirmed that this is correct behavior:

> If we let liquidators liquidate from below liquidation MCR to just above maintenance MCR, if a liquidator only liquidates a small amount, the position will no longer be in liquidation, nor in the healthy zone, so the liquidators can't continue to liquidate.

common/src/borrow.rs
```diff
@@ -137,70 +147,103 @@ impl BorrowPosition {
...
+            .unwrap();
+        let scaled_collateral_amount =
+            collateral_amount * (Decimal::ONE - liquidator_spread) / mcr_liquidation;
+
+        if scaled_liability <= scaled_collateral_amount {
+            CollateralAssetAmount::zero()
+        } else {
+            // Multiplication by mcr_liquidation here could cause overflow
+            let unscaled_amount =
+                (scaled_liability - scaled_collateral_amount) / (mcr_liquidation - Decimal::ONE);
```
vitaliybezz: I see we pass here borrow_mcr_maintenance and borrow_mcr_liquidation values, which are validated in MarketConfiguration::validate as

```        
if self.borrow_mcr_maintenance < 1u32
            || self.borrow_mcr_maintenance < self.borrow_mcr_liquidation
        {
            return Err(error::out_of_bounds("borrow_mcr_maintenance"));
        }

        if self.borrow_mcr_liquidation < 1u32 {
            return Err(error::out_of_bounds("borrow_mcr_liquidation"));
        }
```
Should we change validation to <=   1u32 to avoid possible division by zero?

vitaliybezz (approved):
Looks good to me,  one question.

common/src/borrow.rs
peer2f00l: Good call, changed to <=

vitaliybezz (approved):
I don't see any issues

PR #222, opened 2025-09-16
Make snapshot creation safer & simpler

peer2f00l:
- Require proofs that a snapshot has been taken this block before position guards are accessible.
- Simplify the `snapshot()` function:
    - `current_snapshot` is no longer a long-running, constantly-updated field. Instead, it is dynamically calculated only once: when the snapshot is finalized.
    - Because we don't have to update `current_snapshot` anymore, calling `snapshot()` when a snapshot does not need to be finalized is basically a no-op.
- Structure of market's incoming funds now parallels (using similar logic and same data structures as) supply positions' incoming funds.


vitaliybezz (approved):

common/src/market/impl.rs
```diff
@@ -114,54 +118,80 @@ impl Market {
...
+            index: current_snapshot_index,
+            snapshot: snapshot.clone(),
+        }
+        .emit();
+        self.finalized_snapshots.push(snapshot);
+
+        // Activate incoming funds
+        for i in 0..self.borrow_asset_deposited_incoming.len() {
+            let incoming = &self.borrow_asset_deposited_incoming[i];
+            if incoming.activate_at_snapshot_index == current_snapshot_index {
```
vitaliybezz, line 174: This code activate only the first matching deposit, then exit loop. Is it intentional?

common/src/market/impl.rs
```diff
@@ -35,8 +38,8 @@ pub struct Market {
     pub configuration: MarketConfiguration,
     /// Total amount of borrow asset earning interest in the market.
     pub borrow_asset_deposited_active: BorrowAssetAmount,
-    /// Mapping of upcoming snapshot indices to amounts of borrow asset that will be activated.
-    pub borrow_asset_deposited_incoming: HashMap<u32, BorrowAssetAmount>,
+    /// Upcoming snapshot indices with amounts of borrow asset that will be activated.
+    pub borrow_asset_deposited_incoming: Vec<IncomingDeposit>,
```
vitaliybezz, line 42: What is the benefit to switching from HashMap to Vec?

vitaliybezz (approved):
Couple questions.

common/src/market/impl.rs
peer2f00l: Saving gas on a simpler structure. The collection shouldn't ever be more than 2-3 items long, so all of the overhead of a `HashMap` isn't necessary.

common/src/market/impl.rs
peer2f00l: Yes, this loop shouldn't run more than 1-2 iterations in the vast majority of cases.

PR #228, opened 2025-09-23
Remove fractional snapshot calculations

peer2f00l:
Significantly simplifies code and reduces edge-cases.

### DIFF
--- a/common/src/borrow.rs
+++ b/common/src/borrow.rs
@@ -10,7 +10,7 @@ use crate::{
     market::{Market, SnapshotProof},
     number::Decimal,
     price::{Appraise, Convert, PricePair, Valuation},
-    MS_IN_A_YEAR,
+    MS_PER_YEAR,
 };
 
 /// This struct can only be constructed after accumulating interest on a
@@ -293,7 +293,7 @@ impl<M: Deref<Target = Market>> BorrowPositionRef<M> {
         let interest_in_current_snapshot = self.market.interest_rate()
             * (env::block_timestamp_ms().saturating_sub(prev_end_timestamp_ms))
             * Decimal::from(self.position.get_borrow_asset_principal())
-            / *MS_IN_A_YEAR;
+            / MS_PER_YEAR;
         #[allow(clippy::unwrap_used, reason = "Interest rate guaranteed <= APY_LIMIT")]
         interest_in_current_snapshot.to_u128_ceil().unwrap().into()
     }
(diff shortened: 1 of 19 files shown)


contract/market/tests/snapshot.rs
```diff
@@ -278,17 +268,15 @@ async fn snapshot_with_full_repayment() {
     eprintln!("Total liability before repayment: {total_liability:?}");
 
     // Repay everything (including any accrued interest)
-    c.repay(&borrow_user, total_liability).await;
+    let r = c.repay(&borrow_user, total_liability).await;
+    print_execution(&r);
```
royalf00l: are the prints still intentional?

contract/market/tests/snapshot.rs
```diff
@@ -301,6 +289,9 @@ async fn snapshot_with_full_repayment() {
         final_position.get_total_borrow_asset_liability()
     );
 
+    eprintln!("Final snapshot:");
```
royalf00l, line 291: do we still want these 2 prints?

royalf00l: I guess these are tests, so not a big deal, just double checking

common/src/borrow.rs
```diff
@@ -475,52 +493,117 @@ impl<'a> BorrowPositionGuard<'a> {
...
+            //  more research to see if it is possible for the signature to
+            //  still show up on chain after the promise expires.
+            //
+            // Mitigation strategy: Retain locks until we know the
+            // signature will not be issued. Note that we can't implement
+            // this strategy until we implement asset transfer for MPC
+            // assets, so we IGNORE THIS CASE FOR NOW.
+            //
+            // TODO: Implement case 2 mitigation.
+            // NOTE: Not needed for chain-local (NEP-141-only) tokens.
```
royalf00l, line 605: Empty else case. No code implemented here

peer2f00l: It's here for the comment, which I think is important to acknowledge.

contract/market/tests/snapshot.rs
peer2f00l: this one is unnecessary to keep around

contract/market/tests/snapshot.rs
peer2f00l: yeah I think these ones are ok to keep

royalf00l (approved):

PR #233, opened 2025-09-29
Borrow range max race condition

peer2f00l:


royalf00l (commented):
I don't see a test that confirms this new fix properly solves the race condition problem

royalf00l (approved):

PR #237, opened 2025-09-30
Distribute interest over time instead of all at once

peer2f00l:
Changes:
- All functions related to static yield have been changed: `get_static_yield`, `accumulate_static_yield`, `withdraw_static_yield`
- Borrow interface: `borrow_asset_fees` has been removed and replaced with two fields:
    - `interest` looks and operates almost the same as `borrow_asset_fees`, with the exception that it now _only_ contains the value of the accumulated interest (not one-time fees).
    - `fees` is just a numerical field that contains the amount of one-time fees.
- `GlobalYieldDistributed` event has been removed.
- Supply yield is now earned by calculating the amount of interest that _will be paid_ by the borrowers when they repay their positions. Fees are also distributed when they are _incurred_, instead of when they are _paid_.


common/src/lib.rs
```diff
@@ -18,6 +18,15 @@ pub mod withdrawal_queue;
 
 pub static MS_PER_YEAR: number::Decimal = number::Decimal::from_u64(31_556_952_000); // 1000 * 60 * 60 * 24 * 365.2425
```
royalf00l: Is this still used anywhere or should it be deleted? Seems most of the math seems updated to use `YEAR_PER_MS`

common/src/market/configuration.rs
```diff
@@ -312,7 +313,19 @@ impl MarketConfiguration {
     pub fn single_snapshot_maximum_interest(&self) -> Decimal {
         self.borrow_interest_rate_strategy.at(Decimal::ONE)
             * self.time_chunk_configuration.duration_ms.0
-            / MS_PER_YEAR
+            * YEAR_PER_MS
+    }
+
+    pub fn yield_rate(&self, snapshot: &Snapshot) -> Decimal {
```
royalf00l: consider renaming since this function only calculates the yield from interest, not from liquidation or origination

royalf00l (approved):

SmolinPavel (approved):

PR #239, opened 2025-10-03
chore: Add permissive CORS to relayer server

### LINKED ISSUE #231: Add CORS support to relayer
Frontend needs this.

petarvujovic98:
Add `tower` and `tower-http` crates to support a `CorsLayer` for `axum`. This allows us to use a permissive config to allow easier browser integration.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,7 +44,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom",
+ "getrandom 0.2.15",
  "once_cell",
  "version_check",
 ]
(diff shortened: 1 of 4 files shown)


peer2f00l (commented):
Question: although the relayer is already pretty picky about what sorts of payloads it will actually touch, is there a security risk with permissive CORS vs a [configurable allowed origin list](https://docs.rs/tower-http/latest/tower_http/cors/struct.CorsLayer.html#method.allow_origin)?

petarvujovic98:
Technically there is a higher chance of a DoS attack. From my experience it's just been more of an inconvenience having explicitly listed domains instead of allowing any domain to send requests (makes it easier to deploy different websites for testing or simply having different domains without also changing the backend to add support for them). But from practice I've never actually had a service get DoS attacked because of CORS settings being permissive.
Not sure what the setup is for the relayer infra, but usually having a rate limit in front of the service takes care of most of this.
Also, the origin header is easily manipulated when you are not sending the request from a browser environment so CORS only ever protects you from other websites sending requests in reality.

peer2f00l (approved):

PR #243, opened 2025-10-08
feat: metavault

carrion256:
closes https://linear.app/templar-protocol/issue/ENG-19/initial-vault-contract

Creates https://linear.app/templar-protocol/issue/ENG-39/recallreallocate

TODO:
- [x] use returnstyle
- [x] gas problems with withdraw
- [x] rephrase withdraw>create, exec>exec

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/243)


contract/vault/README.md
```diff
@@ -0,0 +1,61 @@
+# Templar Vault – Withdrawals
+
+This document explains how withdrawals work in the vault as currently implemented.
```
carrion256: Lies now - updated to use a queue ap proach

common/src/vault.rs
```diff
@@ -0,0 +1,486 @@
...
+    //    Under-provisioned gas leads to partial allocations and extra callbacks.
+    //  • Timing control: depositors implicitly decide when allocation runs, which can fight the allocator’s planned rebalancing
+    //    cadence.
+    //  • Thrashing: many small deposits can trigger many allocation passes.
+    //  • Current code is “eager-ish but incomplete”: it only auto-starts when Idle, and does not auto-restart after the op. Deposits
+    //    that arrive during an allocation stay idle until someone triggers another pass.
+    //
+    // Behaviour
+    // • On deposit: if Idle and idle_balance ≥ min_batch, start_allocation(idle_balance).
+    // • Eager allocation can still honor a per-op plan if one is set (plan wins); otherwise fall back to supply_queue order.
```
peer2f00l: Doc comments `///`

common/src/vault.rs
```diff
@@ -0,0 +1,486 @@
+use near_sdk::{env, json_types::U128, near, require, AccountId, Gas, Promise, PromiseOrValue};
+
+use crate::asset::{BorrowAsset, FungibleAsset};
+
+pub type TimestampNs = u64;
+
+pub const ONE_YOCTO: u128 = 1;
+
+pub const MIN_TIMELOCK_NS: u64 = 0;
+pub const MAX_TIMELOCK_NS: u64 = 30 * 86_400_000_000_000; // 30 days
```
peer2f00l, line 14: Should these be configurable?

common/src/vault.rs
```diff
@@ -0,0 +1,486 @@
...
+#[near(serializers = [json])]
+pub enum DepositMsg {
+    /// Add the attached tokens to the sender's vault position.
+    Supply,
+}
+
+#[derive(Clone, Default)]
+#[near]
+pub struct MarketConfiguration {
+    // Supply cap for this market (in underlying asset units)
```
peer2f00l: Doc comments here too. Ditto for any remaining locations that are using normal instead of doc comments.

common/src/vault.rs
```diff
@@ -0,0 +1,486 @@
...
+    pub cap: u128,
+    // Whether market is enabled for deposits/withdrawals
+    pub enabled: bool,
+    // Timestamp (ns) after which market can be removed (if pending removal)
+    pub removable_at: TimestampNs,
+}
+
+impl MarketConfiguration {
+    pub const fn encoded_size() -> usize {
```
peer2f00l, line 72: Add a test to verify this.

common/src/vault.rs
```diff
@@ -0,0 +1,486 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[near(serializers = [json, borsh])]
+/// Operation state machine for asynchronous allocation, withdrawal, and payout flows.
+pub enum OpState {
+    Idle,
+    Allocating {
+        op_id: u64,
+        index: u32,
+        remaining: u128,
```
peer2f00l: `u64` & `u128` should be serialized as strings (`U64` & `U128`) because this struct serializes to/from JSON.

common/src/vault.rs
```diff
@@ -0,0 +1,486 @@
...
+}
+
+#[derive(Clone, Debug)]
+#[near(serializers = [json, borsh])]
+pub struct PendingWithdrawal {
+    pub owner: AccountId,
+    pub receiver: AccountId,
+    pub escrow_shares: u128,
+    pub expected_assets: u128,
+    pub requested_at: u64,
```
peer2f00l, line 432: Ditto for JSON/number serialization.

common/src/vault.rs
```diff
@@ -0,0 +1,486 @@
...
+}
+
+impl PendingWithdrawal {
+    pub const fn encoded_size() -> u64 {
```
peer2f00l: Also test.

contract/vault/examples/gas_report.rs
```diff
@@ -0,0 +1,116 @@
...
+    let max = c.borrow_asset.balance_of(user1.id()).await;
+    let g = || rand::thread_rng().gen_range(0..=max);
+
+    let weights = vec![(c.market.contract().id().clone(), U128(1))];
+    let user1_amount = max / ITERATIONS as u128;
+
+    // Run supplies concurrently.
+    let supply_futures = (0..ITERATIONS).map(|_| async {
+        vault
+            .supply(&user1, user1_amount)
```
peer2f00l, line 32: Running the same user in multiple tasks can cause nonce sync issue with `near-workspaces`.

contract/vault/src/aux.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub fn serialize(
+        &self,
+        amount: templar_common::asset::FungibleAssetAmount<impl templar_common::asset::AssetClass>,
+    ) -> serde_json::Value {
+        match self {
+            Self::Nep141FtTransferCall => serde_json::json!(amount),
+            Self::Nep245MtTransferCall => serde_json::json!([amount]),
+        }
+    }
+}
```
peer2f00l: I think this type exists in the market contract. Could be worth extracting from there -> common.

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1505 @@
...
+pub mod impl_token_receiver;
+pub mod storage_management;
+pub mod wad;
+
+#[cfg(test)]
+mod test_utils;
+
+#[derive(Debug, Clone)]
+#[near(serializers = [json, borsh])]
+#[derive(BorshStorageKey)]
```
peer2f00l, line 56:
```suggestion
#[derive(Debug, Clone, BorshStorageKey)]
#[near(serializers = [borsh])]
```

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1505 @@
...
+    // Pending withdrawals queue (vault-level, FIFO by id)
+    pending_withdrawals: IterableMap<u64, PendingWithdrawal>,
+    next_withdraw_id: u64,
+    next_withdraw_to_execute: u64,
+}
+
+#[near]
+impl Contract {
+    #[allow(clippy::unwrap_used, reason = "Infallible")]
+    #[allow(clippy::too_many_arguments, reason = "Constructor")]
```
peer2f00l: necessary?

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1505 @@
...
+            fee_recipient,
+            skim_recipient,
+            name,
+            symbol,
+            decimals,
+            mode,
+        } = configuration;
+
+        let timelock_ns = u64::from(initial_timelock_sec) * 1_000_000_000;
+        assert!(
```
peer2f00l:
```suggestion
        require!(
```

should include fewer debug symbols

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1505 @@
...
+                self.config
+                    .insert(market.clone(), MarketConfiguration::default());
+                Event::MarketCreated {
+                    market: market.clone(),
+                }
+                .emit();
+                // Pre-allocate a market_supply record (principal=0) so allocations don't create storage later
+                self.market_supply.insert(market.clone(), 0);
+                #[allow(clippy::unwrap_used, reason = "No side effects")]
+                self.config.get_mut(&market).unwrap()
```
peer2f00l:
```suggestion
                self.config.get_mut(&market).unwrap_or_else(|e| env::panic_str(&e.to_string()))
```

Should save on debug symbols

common/src/vault.rs
```diff
@@ -0,0 +1,486 @@
...
+pub enum DepositMsg {
+    /// Add the attached tokens to the sender's vault position.
+    Supply,
+}
+
+#[derive(Clone, Default)]
+#[near]
+pub struct MarketConfiguration {
+    // Supply cap for this market (in underlying asset units)
+    pub cap: u128,
```
peer2f00l: I know the API kind of sucks (will update soon), but this is what the `FungibleAssetAmount` type is for.

common/src/vault.rs
```diff
@@ -0,0 +1,488 @@
...
+    fn preview_withdraw(assets: U128) -> U128;
+    fn preview_redeem(shares: U128) -> U128;
+}
+
+// FIXME: move to market
+pub const GET_SUPPLY_POSITION_GAS: Gas = Gas::from_tgas(4);
+// FIXME: move to market
+pub const CREATE_WITHDRAW_REQ_GAS: Gas = Gas::from_tgas(10);
+// FIXME: move to market
+pub const EXECUTE_WITHDRAW_REQ_GAS: Gas = Gas::from_tgas(10);
```
carrion256: Assuming these exist somewhere akin to L155-L159

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1505 @@
...
+    /// Sets the ordered supply (allocation) queue.
+    /// Rejects duplicates and markets without a positive cap. Requires the vault to be idle.
+    #[payable]
+    pub fn set_supply_queue(&mut self, markets: Vec<AccountId>) {
+        Self::assert_allocator();
+        self.ensure_idle();
+        assert!(markets.len() <= MAX_QUEUE_LEN, "too long");
+
+        // Invariant: supply_queue has no duplicates; allocation order remains meaningful
+        let mut seen = std::collections::HashSet::new();
```
peer2f00l:
```suggestion
        let mut seen = std::collections::HashSet::with_capacity(markets.len());
```

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1505 @@
...
+    #[payable]
+    pub fn set_withdraw_queue(&mut self, queue: Vec<AccountId>) {
+        Self::assert_allocator();
+        self.ensure_idle();
+        assert!(
+            queue.len() <= MAX_QUEUE_LEN,
+            "Withdraw queue length exceeds max"
+        );
+
+        let mut seen = std::collections::HashSet::new();
```
peer2f00l:
```suggestion
        let mut seen = std::collections::HashSet::with_capacity(queue.len());
```

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1505 @@
...
+                "Invariant violation: Unknown market in new queue"
+            );
+        }
+
+        for (id, cfg) in self.config.iter() {
+            let has_supply = *self.market_supply.get(id).unwrap_or(&0) > 0;
+            println!(
+                "ID: {}, Enabled: {}, Has Supply: {}, Removable At: {}",
+                id, cfg.enabled, has_supply, cfg.removable_at
+            );
```
peer2f00l:
```suggestion
```

Or `#[cfg]`-gate, or `near_sdk::log!`

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+
+            if (cfg.enabled || has_supply) && !seen.contains(id) {
+                if current.contains(id) {
+                    // Omission is allowed only when removing an existing queued market AND all safety preconditions hold.
+                    assert!(
+                        cfg.cap == 0,
+                        "Invariant violation: Cannot remove market with non-zero cap"
```
peer2f00l: Invariants are things that are always true about the contract's state; an "Invariant violation" message should never appear if a contract is operating normally (i.e. within the assumptions the programmer made during development).

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+            self.start_allocation(total)
+        } else {
+            // Validate unique markets and accumulate weight sum
+            let mut seen = std::collections::HashSet::new();
+            let mut sum_w: u128 = 0;
+
+            for (m, w) in &weights {
+                if !seen.insert(m.clone()) {
+                    env::panic_str(&format!("Duplicate market in weights: {m}"));
+                }
```
peer2f00l: Instead of this pattern that I keep seeing, you could simply `.iter().collect::<HashSet<_>>()` the input `Vec`s and then compare lengths. That might be simpler than explicitly writing loops like this.

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+}
+
+/* ----- Views ----- */
+#[near]
+impl Contract {
+    #[allow(clippy::expect_used, reason = "No side effects")]
+    pub fn get_configuration(&self) -> VaultConfiguration {
+        let timelock_sec = self.timelock_ns / 1_000_000_000;
+        VaultConfiguration {
+            owner: self.own_get_owner().expect("Owner not set"),
```
peer2f00l: Same with `expect` - I think it carries a higher cost of debug symbols.

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1506 @@
...
+    pub(crate) fn clamp_allocation_total(&self, requested: Option<u128>) -> u128 {
+        let requested = requested.unwrap_or(self.idle_balance);
+        let max_room = self.get_max_deposit().0;
+        requested.min(self.idle_balance).min(max_room)
+    }
+
+    pub(crate) fn compute_escrow_settlement(
+        escrow_shares: u128,
+        burn_shares: u128,
+    ) -> (u128 /* to_burn */, u128 /* refund */) {
```
peer2f00l: Return a struct so the values are labeled.

contract/vault/src/wad.rs
```diff
@@ -0,0 +1,130 @@
+/// Fixed-point helpers and fee-accrual math using 18-decimal WAD precision.
```
peer2f00l: We do have `templar_common::number::Decimal` which covers much of the same ground as this lib.

test-utils/src/lib.rs
```diff
@@ -83,14 +85,43 @@ macro_rules! accounts {
...
+            $(s.vault.init_account(&$n)),*
         );
         let $crate::SetupEverything { $($e,)* .. } = s;
     };
     ($w:ident extract($($e:ident),*) accounts($($n:ident),*)) => {
-        $crate::setup_test!($w extract($($e),*) accounts($($n),*) config(|_| {}))
+        $crate::setup_test!($w extract($($e),*) accounts($($n),*) config(|_| {}) vconfig(|_| {}))
+    };
+    (extract($($e:ident),*) accounts($($n:ident),*) config($f:expr)) => {
+        let worker = near_workspaces::sandbox().await.unwrap();
```
peer2f00l: Prefer the `#[fixture]` to reduce the number of times we call `near_workspaces::sandbox()` and subsequently dispose of it.

Cargo.toml
```diff
@@ -32,6 +33,7 @@ rstest = { version = "0.24" }
 schemars = { version = "0.8" }
 templar-common = { path = "./common" }
 templar-universal-account = { path = "./universal-account" }
+template-vault-contract = { path = "./contract/vault" }
```
peer2f00l:
```suggestion
templar-vault-contract = { path = "./contract/vault" }
```

peer2f00l (commented):
First look. Not comprehensive, and most comments are pretty surface-level. I'm not confident that I have a complete-picture understanding of everything that's going on yet, so we should hop on a call and run through this a few times before we call it good.

common/src/vault.rs
carrion256: I think this is a large bound, I couldnt imagine anything going beyond 30 days for a cap.

common/src/vault.rs
carrion256: I think the U128 only applies to bignums, are you saying every json-serialisable type should use the string format?

EDIT: added some

contract/vault/src/lib.rs
carrion256: I belive this is equivalent and not constrained to contracts: it is a protocol invariant that you cannot remove a market which has a non-zero cap.

common/src/vault.rs
carrion256: I would prefer to defer to the API when it is easier to use in this case.

peer2f00l: on it

contract/vault/src/wad.rs
carrion256: This type is not so useful for share conversion since we don't do any pow or exp. Only pessimistic accounting is required for the vault(redeem/mint), which is simpler, exact and more efficient. 
Furthermore, the rest of DeFi does WAD based accounting, so it's more auditable and exact.

contract/vault/src/wad.rs
```diff
@@ -0,0 +1,445 @@
...
+    fn from(v: u128) -> Self {
+        Number(U256::from(v))
+    }
+}
+impl From<Number> for u128 {
+    #[inline]
+    fn from(n: Number) -> u128 {
+        n.as_u128_trunc()
+    }
+}
```
peer2f00l, line 105: Should these two implementations not be inverses of each other?

common/src/vault.rs
```diff
@@ -0,0 +1,604 @@
...
+    fn after_send_to_user(&mut self, op_id: u64, receiver: AccountId, amount: U128) -> bool;
+    fn after_skim_balance(&mut self, token: AccountId, recipient: AccountId) -> bool;
+}
+
+#[derive(Clone)]
+#[near]
+pub struct PendingValue<T> {
+    pub value: T,
+    // Timestamp when this pending value can be finalized
+    pub valid_at: TimestampNs,
```
peer2f00l: Consider renaming to `valid_at_ns`, so that when the struct is serialized to JSON the unit information is not lost.

contract/vault/src/impl_token_receiver.rs
```diff
@@ -0,0 +1,156 @@
...
+            return deposit;
+        };
+
+        if deposit == 0 {
+            Event::DepositRejectedZeroAmount {
+                sender: sender_id.clone(),
+            }
+            .emit();
+            return 0;
+        }
```
peer2f00l: I don't think that there's a _strong_ convention here one way or another, but the market contract simply rejects invalid transactions. Failed `*_on_transfer` calls get refunded, plus explorers, etc. will have slightly better visibility into which receipts failed, because they can rely on protocol-level failures instead of custom events.

contract/vault/src/impl_token_receiver.rs
```diff
@@ -0,0 +1,156 @@
...
+
+        self.internal_accrue_fee();
+
+        let max = self.get_max_deposit().0;
+        let accept = deposit.min(max);
+        let refund = deposit - accept;
+
+        let shares = self.preview_deposit(U128(accept)).0;
```
peer2f00l, line 123: Is the reason that we don't simply call `convert_to_shares` directly for semantics?

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1597 @@
...
+        let refund = escrow_shares.saturating_sub(to_burn);
+        EscrowSettlement { to_burn, refund }
+    }
+
+    /* ----- Internal: fee, shares ----- */
+    pub fn mint_shares(&mut self, to: &AccountId, amount: u128) {
+        if amount == 0 {
+            return;
+        }
+        #[allow(clippy::expect_used, reason = "No side effects")]
```
peer2f00l: Unnecessary

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1597 @@
...
+        self.deposit_unchecked(to, amount)
```
peer2f00l: Use `.mint` for NEP-141 standard event emission.

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1597 @@
...
+                self.op_state
+            ));
+        }
+    }
+
+    fn start_allocation(&mut self, amount: u128) -> PromiseOrValue<()> {
+        if amount == 0 {
+            return self.stop_and_exit(Some(&Error::ZeroAmount));
+        }
+        self.ensure_idle();
```
peer2f00l, line 823: All non-test invocations of `start_allocation` are preceded by an idle check, so this might be redundant.

contract/vault/src/impl_callbacks.rs
```diff
@@ -0,0 +1,717 @@
...
+        match self.op_state {
+            OpState::Allocating { .. } => self.stop_and_exit_allocating(msg),
+            OpState::Withdrawing { .. } => self.stop_and_exit_withdrawing(msg),
+            OpState::Payout { .. } => self.stop_and_exit_payout(msg),
+            OpState::Idle => {
+                Event::OperationStoppedWhileIdle {
+                    reason: msg.map(std::string::ToString::to_string),
+                }
+                .emit();
+                self.op_state = OpState::Idle;
```
peer2f00l: Redundant?

common/src/vault.rs
```diff
@@ -0,0 +1,604 @@
...
+    /// - On completion of allocation: Withdrawing (to satisfy pending user requests) or Idle (if stopped).
+    /// - On stop/failure: Idle.
+    Allocating {
+        /// Unique operation id used to correlate async callbacks and detect drift.
+        op_id: u64,
+        /// Zero-based position within the allocation plan/queue currently being processed.
+        index: u32,
+        /// Amount of underlying (in asset units) still to allocate during this operation.
+        remaining: u128,
+    },
```
peer2f00l: Use named structs for each variant so that we can destructure in places like [here](https://github.com/Templar-Protocol/contracts/blob/8b13f47ed769f08971ec9da31e9f1b4ce0b34eba/contract/vault/src/impl_callbacks.rs#L597) and avoid `if let` [here](https://github.com/Templar-Protocol/contracts/blob/8b13f47ed769f08971ec9da31e9f1b4ce0b34eba/contract/vault/src/impl_callbacks.rs#L482-L486)

contract/vault/src/impl_callbacks.rs
```diff
@@ -0,0 +1,717 @@
...
+                        remaining: U128(*remaining),
+                        reason: Some(m.to_string()),
+                    }
+                    .emit();
+                }
+            }
+
+            if *remaining > 0 {
+                self.idle_balance = self.idle_balance.saturating_add(*remaining);
+            }
```
peer2f00l:
```suggestion
            msg.map_or(Event::AllocationCompleted { op_id: *op_id }, |m| {
                Event::AllocationStopped {
                    op_id: (*op_id).into(),
                    index: *index,
                    remaining: U128(*remaining),
                    reason: Some(m.to_string()),
                }
            })
            .emit();

            self.idle_balance = self.idle_balance.saturating_add(*remaining);
```

contract/vault/src/impl_callbacks.rs
```diff
@@ -0,0 +1,717 @@
...
+                owner,
+                escrow_shares,
+                ..
+            } if *escrow_shares > 0 => Some((owner.clone(), *escrow_shares)),
+            _ => None,
+        } {
+            let self_id = env::current_account_id();
+            #[allow(clippy::expect_used, reason = "No side effects")]
+            self.transfer_unchecked(&self_id, &owner_acc, escrow)
+                .unwrap_or_else(|e| env::log_str(&e.to_string()));
```
peer2f00l, line 595: Use `transfer` for proper event emission, also the `#[allow]` is no longer needed.

contract/vault/src/impl_callbacks.rs
```diff
@@ -0,0 +1,717 @@
...
+            ..
+        } = &self.op_state
+        {
+            if *escrow_shares > 0 {
+                let self_id = env::current_account_id();
+                let owner_acc = owner.clone();
+                let escrow = *escrow_shares;
+                #[allow(clippy::expect_used, reason = "No side effects")]
+                self.transfer_unchecked(&self_id, &owner_acc, escrow)
+                    .unwrap_or_else(|e| env::log_str(&e.to_string()));
```
peer2f00l: Ditto with above

contract/vault/src/impl_callbacks.rs
```diff
@@ -0,0 +1,717 @@
...
+                    reason: msg.map(std::string::ToString::to_string),
+                }
+                .emit();
+            }
+        }
+        if let OpState::Payout {
+            owner,
+            escrow_shares,
+            ..
+        } = &self.op_state
```
peer2f00l: Why are we `if let` destructuring twice?

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1594 @@
...
+
+    /// Accepts the pending Guardian change after the timelock has elapsed.
+    pub fn accept_guardian(&mut self) {
+        Self::require_owner();
+
+        let p = self.pending_guardian.clone();
+
+        if let Some(p) = &p {
+            require!(env::block_timestamp() >= p.valid_at, "not yet");
+            Self::with_members_of(&Role::Guardian, |members| {
```
peer2f00l: Use `with_members_of_mut` to manipulate the membership set directly (no `mut`-borrow collisions).

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1594 @@
...
+    pub fn set_performance_fee(&mut self, fee: U128) {
+        Self::require_owner();
+
+        let fee_wad = wad::Wad::from(fee.0);
+
+        require!(
+            fee_wad != self.performance_fee,
+            "Fee already set to this value"
+        );
+        require!(fee_wad <= (wad::Wad::one() / 10), "fee too high");
```
peer2f00l: Make configurable or named constant.

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1594 @@
...
+            account: account.clone(),
+        }
+        .emit();
+        self.storage_deposit(Some(account.clone()), Some(true));
+
+        self.fee_recipient = account;
+    }
+
+    /// Sets the performance fee as a WAD fraction (1e24 = 100%). Accrues fees at the old rate first.
+    pub fn set_performance_fee(&mut self, fee: U128) {
```
peer2f00l: Can `Wad` implement `serde::Deserialize` directly?

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1594 @@
...
+    }
+
+    /* ----- Timelocks / Pending ----- */
+    /// Proposes a new governance timelock in nanoseconds.
+    /// If increasing, applies immediately; if decreasing, starts a timelock equal to the current duration.
+    pub fn submit_timelock(&mut self, new_timelock_ns: U64) {
+        Self::require_owner();
+        let tl = &new_timelock_ns.0;
+
+        require!(tl != &self.timelock_ns, "Already set to this value");
```
peer2f00l: Idempotence?

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1594 @@
...
+    }
+
+    /// Accepts a pending timelock change after it becomes valid.
+    pub fn accept_timelock(&mut self) {
+        Self::require_owner();
+        if let Some(p) = &self.pending_timelock {
+            require!(
+                env::block_timestamp() >= p.valid_at,
+                "Timelock not elapsed yet"
+            );
```
peer2f00l: This kind of check appears a few different places in the code: can be extracted to a method on `PendingValue`?

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1594 @@
...
+    pending_timelock: Option<PendingValue<TimestampNs>>,
+    /// Any pending change to the vault's guardian
+    pending_guardian: Option<PendingValue<AccountId>>,
+    /// Current timelock duration for governance actions (ns)
+    timelock_ns: TimestampNs,
+
+    /// Ordered list of market IDs for deposit allocation
+    supply_queue: Vector<AccountId>,
+    /// Ordered list of market IDs for withdrawal prioritytr
+    withdraw_queue: Vector<AccountId>,
```
peer2f00l: If the number of markets is fairly low + constant, it is probably not gas-efficient to use the `near_sdk` collections here - in-memory `Vec` is probably much faster + cheaper, possibly wrapped in `near_sdk::store::Lazy`

contract/vault/src/lib.rs
```diff
@@ -0,0 +1,1594 @@
...
+
+    /// The process in which the vault calculates its assets under management (AUM)
+    aum: AUM,
+
+    /// The mode in which the allocator will operate
+    mode: AllocationMode,
+    plan: Option<AllocationPlan>,
+
+    /// configuration per market (market ID -> MarketConfig)
+    markets: IterableMap<AccountId, MarketConfiguration>,
```
peer2f00l: Could `markets`, `pending_cap`, and `market_supply` be combined into a single `IterableMap`?

carrion256: Good question - I think markets & market supply, possibly pending cap. Will take a look

contract/vault/src/impl_token_receiver.rs
carrion256: This is mostly an ethism, but the main use case is to maintain compatibility (on a like-for-like basis) with erc4626. Some vaults override the function for preview_shares to introduce any fee mechanics or such.

contract/vault/src/impl_callbacks.rs
carrion256: Unfortunately we need to do a best-effort here. If the transfer panics then we will not reconcile the state (for inflight withdrawals, payouts, etc). In this case, if the user forces a panic by withdrawing from storage (when we have storage management) then the withdrawal queue would be bricked.

I think this is a design consideration for storage management (proper) vs the approach i have used thus far. If we do intend to eventually implement proper storage management (in the eyes of NEAR) then we have to consider extra edge cases like above.

Edit: transfer also does not panic

contract/vault/src/impl_callbacks.rs
carrion256: See above

contract/vault/src/lib.rs
carrion256: Yep - may aswell panic rather than set it again

contract/vault/src/lib.rs
carrion256: I agree with this - there can be at most 20 markets, so let's put this in memory to mitigate storage combinations

peer2f00l (approved):

PR #244, opened 2025-10-09
fix: Fetch markets from registries and add service file

petarvujovic98:
Adds a service file for the accumulator bot. Updates the implementation to fetch markets from a list of registries and moves some code to be reused between the liquidator and accumulator bots.

### DIFF
--- a/bots/accumulator.service
+++ b/bots/accumulator.service
@@ -1,0 +1,14 @@
+; vim: set filetype=dosini:
+[Unit]
+Description=Templar Accumulator Bot
+After=network.target
+
+[Service]
+EnvironmentFile=/etc/default/accumulator
+ExecStart="/usr/local/bin/accumulator"
+Restart=on-failure
+User=root
+WorkingDirectory=/usr/local/bin
+
+[Install]
+WantedBy=multi-user.target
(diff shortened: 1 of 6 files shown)


bots/accumulator.service
```diff
@@ -0,0 +1,14 @@
+; vim: set filetype=dosini:
+[Unit]
+Description=Templar Accumulator Bot
+After=network.target
+
+[Service]
+EnvironmentFile=/etc/default/accumulator
+ExecStart="/usr/local/bin/accumulator"
+Restart=on-failure
+User=root
```
peer2f00l: Maybe should have its own user?

bots/src/accumulator.rs
```diff
@@ -35,16 +35,40 @@ pub struct Args {
...
+        short = 'r',
+        long,
+        default_value = "3600",
+        env = "REGISTRY_REFRESH_INTERVAL"
+    )]
+    pub registry_refresh_interval: u64,
     /// Concurrency for accumulation tasks
-    #[arg(short, long, default_value = "10", env = "CONCURRENCY")]
+    #[arg(short, long, default_value = "4", env = "CONCURRENCY")]
     pub concurrency: usize,
```
peer2f00l: `default_value` vs `default_value_t`?

peer2f00l (approved):
Nothing too big, likely fine as-is.

bots/accumulator.service
petarvujovic98: Yup, will update this

bots/src/accumulator.rs
petarvujovic98: Ahh, yes, missed that

PR #250, opened 2025-10-14
chore: make migrations more flexible + fmt

peer2f00l:


service/relayer/migrations/20251004090744_spend_within_transaction.down.sql
```diff
@@ -31,15 +31,15 @@ ALTER COLUMN
...
     NOT NULL,
-    DROP COLUMN allowance_spent_inner,
+    DROP COLUMN IF EXISTS allowance_spent_inner,
 ADD
-    COLUMN id uuid NOT NULL DEFAULT gen_random_uuid(),
+    COLUMN IF NOT EXISTS id uuid NOT NULL DEFAULT gen_random_uuid(),
     DROP CONSTRAINT pk__transaction,
 ADD
     CONSTRAINT pk__call PRIMARY KEY (id),
     DROP COLUMN STATUS;
```
vitaliybezz: column name should be lower case "status" for consistency.

vitaliybezz (approved):

service/relayer/migrations/20251004090744_spend_within_transaction.down.sql
peer2f00l: Good catch!

PR #258, opened 2025-10-15
Add codecov coverage setup

vitaliybezz:


royalf00l:
Glad the tests are passing, but I think a few things need to be changed:
1. I don't see the test coverage on https://app.codecov.io/gh/Templar-Protocol/contracts 
2. Does the cargo-llvm-cov (or any of the coverage tools from https://about.codecov.io/language/rust/) only cover inline tests and exclude test coverage from integration tests, specifically ones that happen using the NEAR runtime? If so, I think we should try to use https://github.com/hknio/wasmcov as suggested https://t.me/c/2452747383/2/3396. Even if this is not compatible with codecov.io, you should be able to generate a report on each PR using the wasmcov tool + gh actions
3. the [test / coverage action](https://github.com/Templar-Protocol/contracts/actions/runs/18570029064/job/52973660123?pr=258) took 49 minutes to run. Hopefully switching to wasmcov will speed that up :)

.github/workflows/test.yml
```diff
@@ -42,3 +60,61 @@ jobs:
         env:
           SQLX_OFFLINE: true
         run: ./script/test.sh
+
+  coverage:
+    name: Coverage
+    runs-on: ubuntu-latest
```
royalf00l, line 66: Maybe ubuntu minimal or lubuntu will speed up coverage generation time

vitaliybezz:
> Glad the tests are passing, but I think a few things need to be changed:
> 
> 1. I don't see the test coverage on https://app.codecov.io/gh/Templar-Protocol/contracts
> 2. Does the cargo-llvm-cov (or any of the coverage tools from https://about.codecov.io/language/rust/) only cover inline tests and exclude test coverage from integration tests, specifically ones that happen using the NEAR runtime? If so, I think we should try to use https://github.com/hknio/wasmcov as suggested https://t.me/c/2452747383/2/3396. Even if this is not compatible with codecov.io, you should be able to generate a report on each PR using the wasmcov tool + gh actions
> 3. the [test / coverage action](https://github.com/Templar-Protocol/contracts/actions/runs/18570029064/job/52973660123?pr=258) took 49 minutes to run. Hopefully switching to wasmcov will speed that up :)

1. Coverage will appear after merge, but even now you can select current branch and see coverage report.
2. I experimented with WasmCov, it will require a more complex setup with separate workflow and most likely with docker for consistency. Probably the best and the most reliable approach would be to measure coverage on unit tests only (which is pretty common in industry), and keep more complex integration tests which require some env setup separate. We can build up the right amount of unit tests with target coverage with time. Such setup will be cleaner and more reliable. 
3. It takes longer because of zero optimization to measure coverage correctly. Also for some reason some tests fails and rerun needed. Another reason to run coverage on unit tests only.

royalf00l:
1. I see it now https://app.codecov.io/github/Templar-Protocol/contracts/tree/codecov-coverage 
Screen Shot 2025-10-17 at 2 30 31 PM


2. The problem is that smart contract functions can't be properly tested with simple unit tests, they require a the wasm NEAR contract runtime, which requires https://github.com/hknio/wasmcov to get the coverage for. I understand that in typical backend system unit test coverage is separate from integration test coverage, but the tests in the contract directory all require the wasm NEAR smart contract runtime 

3. Understood

The reason we need to include those tests is bc potential depositors care about the test coverage of the contracts. We have tests for the contracts, so the coverage metrics should reflect those tests. It may be worth highlighting the test coverage of the `contract/` directory (the most important directory which we have test coverage for, but the llvm coverage tool says is not covered) because everything outside of that is an external service

peer2f00l:
After some conversation with @carrion256 , it seems like our best bet might be to convert / duplicate some of the integration tests into traditional Rust `#[test]` tests so that they are covered by the coverage tool, since integrating a tool like wasmcov seems like a _lot_ of extra work.

royalf00l:
> After some conversation with @carrion256 , it seems like our best bet might be to convert / duplicate some of the integration tests into traditional Rust `#[test]` tests so that they are covered by the coverage tool, since integrating a tool like wasmcov seems like a _lot_ of extra work.

OK, I'll approve current PR and created a separate issue to track: https://github.com/Templar-Protocol/contracts/issues/263

royalf00l (approved):

PR #262, opened 2025-10-20
feat: Add initial fuzz targets

petarvujovic98:
This PR adds initial support for code fuzzing using `cargo fuzz`.
Currently the fuzz targets are working off of some assumed flows and should be improved for proper usage.
The limitations of existing fuzzing tooling require us to add some feature flag gating in order to have the code we want to fuzz compile in the fuzz target environment. 
Additionally, some fields have been made public, but those could be reverted with preferred usage patterns without polluting our intended public API.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/262)


peer2f00l (approved):
Is this something that we should add to our Github workflows?

peer2f00l (changes requested):
Ah, I see that the gas report is broken. Just that, I suppose.

fuzz/fuzz_targets/fuzz_borrow.rs
```diff
@@ -0,0 +1,158 @@
...
+        }
+    }
+
+    // Test timestamp handling
+    if timestamp_ms > 0 {
+        position.started_at_block_timestamp_ms = Some(near_sdk::json_types::U64(timestamp_ms));
+    }
+
+    // Test different operations based on selector
+    match op_selector % 8 {
```
carrion256, line 69: I would consider making these their own tests and fuzzing the inputs for each one, in this case we're watering down potential fuzz paths

fuzz/fuzz_targets/fuzz_decimal_arithmetic.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    // Fuzz is_zero
+    let _ = dec_a.is_zero();
+    let _ = Decimal::ZERO.is_zero();
+
+    // Fuzz fractional part
+    let _ = dec_a.fractional_part_as_u128_dividend();
+
+    // Test edge cases based on operation selector
+    match op_selector % 10 {
```
carrion256, line 102: Same as above for these switchlike statements

fuzz/fuzz_targets/fuzz_decimal_parsing.rs
```diff
@@ -0,0 +1,86 @@
...
+                    decimal.near_equal(reparsed),
+                    "Round-trip failed: original={decimal:?}, string={to_string}, reparsed={reparsed:?}",
+                );
+            }
+
+            // Test various precision levels
+            for precision in [0, 1, 5, 10, 20, 38] {
+                let fixed = decimal.to_fixed(precision);
+                let _ = Decimal::from_str(&fixed);
+            }
```
carrion256, line 27: I would fuzz everything up to a known fixed decimal here

fuzz/fuzz_targets/fuzz_decimal_parsing.rs
```diff
@@ -0,0 +1,86 @@
...
+    // Test edge case strings
+    let edge_cases = [
+        "0",
+        "1",
+        "0.0",
+        "1.0",
+        "0.1",
+        "0.00000001",
+        "999999999999999",
+    ];
```
carrion256, line 69: I think also here - fuzzing inputs should be the edge cases and countertests for them

fuzz/fuzz_targets/fuzz_decimal_parsing.rs
```diff
@@ -0,0 +1,86 @@
...
+        if let Ok(decimal) = Decimal::from_str(case) {
+            let stringified = decimal.to_fixed(38);
+            let _ = Decimal::from_str(&stringified);
+        }
+    }
+
+    // Test malformed strings don't panic
+    let malformed = [
+        ".", ".0", "0.", "abc", "1.2.3", "-1", "1e10", "NaN", "Infinity", "",
+    ];
```
carrion256, line 81: Also here, i think probably a few different fuzz tests for these multiple assertions, since finding a negative input would require running the entire suite again

fuzz/fuzz_targets/fuzz_decimals.rs
```diff
@@ -0,0 +1,35 @@
...
+use templar_common::interest_rate_strategy::UsageCurve;
+use templar_common::interest_rate_strategy::{Exponential2, Piecewise};
+use templar_common::number::Decimal;
+
+// Helper to convert u64 to a Decimal in [0, 1]
+fn to_decimal01(x: u64) -> Decimal {
+    Decimal::from(x) / Decimal::from(u64::MAX)
+}
+
+fuzz_target!(|data: (u64, u64, u64, u64, u64, u64, u64)| {
```
carrion256, line 13: Ditto here

fuzz/fuzz_targets/fuzz_interest_math.rs
```diff
@@ -0,0 +1,189 @@
...
+    //             "Compound interest resulted in less than principal"
+    //         );
+    //
+    //         // Invariant: Result shouldn't be absurdly large
+    //         assert!(
+    //             result <= principal.saturating_mul(100),
+    //             "Compound interest grew unreasonably"
+    //         );
+    //     }
+    // }
```
carrion256, line 70: Is this still needed?

carrion256 (commented):
I would consider splitting up a bunch of the giant tests into more composable tests - right now if we meet any type of counter pattern from fuzzing we'd need to fuzz a lot to get to our desired state.

petarvujovic98:
@carrion256 Thanks for the thorough review! I'll update this over the weekend to allow for more granular fuzzing and a faster iteration cycle.

carrion256 (approved):

peer2f00l (approved):

peer2f00l:
All tests pass locally; must be an issue with the runner again.

PR #264, opened 2025-10-21
feat: Supply withdrawal partial fulfillment

peer2f00l:
Semantic change: `execute_next_supply_withdrawal_request` may not always reduce the length of the supply withdrawal queue even if it succeeds, because it will now perform partial fulfillment if possible.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/264)


peer2f00l:
This PR could be further improved to support sending multiple fulfillments in a single `execute_next` call.

For instance, the `WithdrawalQueue` API could be altered to include a function a la:

```rust
fn pop_amount(&mut self, amount: BorrowAssetAmount, limit: u32) -> Vec<(AccountId, BorrowAssetAmount)>;
```

Which would return a list of accounts up to `limit` long that can be fulfilled from the queue given `amount` of available liquidity for withdrawal fulfillments. These could all be fulfilled simultaneously, and should slightly reduce the number of `execute_next` calls necessary to clear the queue, and it should also save a bit on gas.

@carrion256 Comments?

carrion256:
*[`common/src/withdrawal_queue.rs` line 221 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/264#-Oc5fPB4CpkIYqGGAy3d:-Oc5fPB4CpkIYqGGAy3e:b-odl722) ([raw file](https://github.com/Templar-Protocol/contracts/blob/26ecda05848151bc17148fbedd573668ec4f4ef8/common/src/withdrawal_queue.rs#L221)):*
> ```XML
>     pub fn insert_or_update(&mut self, account_id: &AccountId, amount: BorrowAssetAmount) {
>         if let Some(node_id) = self.entries.get(account_id) {
>             if self.is_locked && Some(node_id) == self.queue_head {
> ```

Is this the only place that locking is a requirement?

carrion256:
*[`common/src/market/configuration.rs` line 259 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/264#-Oc5faMD3N9voWvMDHv_:-Oc5faMD3N9voWvMDHva:b8zzm6l) ([raw file](https://github.com/Templar-Protocol/contracts/blob/26ecda05848151bc17148fbedd573668ec4f4ef8/common/src/market/configuration.rs#L259)):*
> ```XML
> 
>         if let Fee::Flat(amount) = self.supply_withdrawal_fee.fee {
>             if amount > self.supply_withdrawal_range.minimum {
> ```

NIT: Would be nice if the error message logged the minimum too

carrion256:
*[`contract/market/src/impl_helper.rs` line 187 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/264#-Oc5fuH-5hNNzSqLf4Ag:-Oc5fuH-5hNNzSqLf4Ah:b-k304y7) ([raw file](https://github.com/Templar-Protocol/contracts/blob/26ecda05848151bc17148fbedd573668ec4f4ef8/contract/market/src/impl_helper.rs#L187)):*
> ```XML
>         let mut supply_position = self
>             .supply_position_guard(snapshot, withdrawal_resolution.account_id.clone())
>             .unwrap_or_else(|| env::panic_str("Invariant violation: Nonexistent supply position"));
> ```

Do we have to sanitise any state from previous cb's here?

carrion256:
*[`contract/market/src/impl_helper.rs` line 189 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/264#-Oc5gKua5lpzXWh9W3YM:-Oc5gKua5lpzXWh9W3YN:b-xzv8jj) ([raw file](https://github.com/Templar-Protocol/contracts/blob/26ecda05848151bc17148fbedd573668ec4f4ef8/contract/market/src/impl_helper.rs#L189)):*
> ```XML
>             .unwrap_or_else(|| env::panic_str("Invariant violation: Nonexistent supply position"));
>         supply_position.record_withdrawal_final(&withdrawal_resolution, withdrawal_succeeded);
>         drop(supply_position);
> ```

NIT: it might be clean to scope the use of this rather than directly invoke drop, but thats just a style suggestion

carrion256 (commented):
One question and a few nits

> @carrion256 reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/264#-:-Oc5gxgJ4FVqrVVTUo04:b22mbt1)* status: all files reviewed, 4 unresolved discussions (waiting on @peer2f00l)

peer2f00l (commented):
> *[Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/264)* status: all files reviewed, 4 unresolved discussions (waiting on @carrion256)

___
*[`common/src/withdrawal_queue.rs` line 221 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/264#-Oc5fPB4CpkIYqGGAy3d:-Oc5nqenEJ7OkP2l5vrq:b-izvmkg) ([raw file](https://github.com/Templar-Protocol/contracts/blob/26ecda05848151bc17148fbedd573668ec4f4ef8/common/src/withdrawal_queue.rs#L221)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

Is this the only place that locking is a requirement?
</blockquote></details>

Locking is to ensure there is not a race condition between initiating a withdrawal and modifying one.

However, if we opt to do batching, then we will pop from the queue eagerly, so there should be no need for locking the queue at all.
___
*[`common/src/market/configuration.rs` line 259 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/264#-Oc5faMD3N9voWvMDHv_:-Oc5oRwCEOBvPwYhEu_N:b-xtfe2f) ([raw file](https://github.com/Templar-Protocol/contracts/blob/26ecda05848151bc17148fbedd573668ec4f4ef8/common/src/market/configuration.rs#L259)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

NIT: Would be nice if the error message logged the minimum too
</blockquote></details>

The minimum is also in the configuration that is being validated.
___
*[`contract/market/src/impl_helper.rs` line 187 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/264#-Oc5fuH-5hNNzSqLf4Ag:-Oc5oe6B0oiih1eT4yqG:bh4qji) ([raw file](https://github.com/Templar-Protocol/contracts/blob/26ecda05848151bc17148fbedd573668ec4f4ef8/contract/market/src/impl_helper.rs#L187)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

Do we have to sanitise any state from previous cb's here?
</blockquote></details>

We check whether the previous callback succeeded, but since it's just a transfer, it doesn't return much meaningful information.
___
*[`contract/market/src/impl_helper.rs` line 189 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/264#-Oc5gKua5lpzXWh9W3YM:-Oc5ol__5uX7zuzv-GLO:bebrt1w) ([raw file](https://github.com/Templar-Protocol/contracts/blob/26ecda05848151bc17148fbedd573668ec4f4ef8/contract/market/src/impl_helper.rs#L189)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

NIT: it might be clean to scope the use of this rather than directly invoke drop, but thats just a style suggestion
</blockquote></details>

Will do

peer2f00l:
Depends on #270

carrion256 (approved):

PR #272, opened 2025-10-27
feat: reallocate

carrion256:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/272)

- [ ] BREAKING CHANGE: Introduces delta-based allocations, `allocate` no longer exists in favour of deltas +Supply/-Withdraw
- [ ] Removes the eager allocation mode for simplicity
- [ ] Adds timeout for workflows since we now have a buncha contract tests
- [ ] Improves vault docs
- [ ] Cleans up some old errors
- [ ] FIX: rebalance withdraw into self to withdraw from a market
- [ ] Improve shallow depth receipt panics
- [ ] Improve testing approaches to make better use of rstest


common/src/vault.rs
```diff
@@ -711,25 +805,105 @@ pub enum Event {
...
+        self.to_lock.contains(&i)
+    }
+
+    pub fn is_locked_all(&self) -> bool {
+        !self.to_lock.is_empty()
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
carrion256: remove these

carrion256:
@copilot describe the changes in this pr

contract/vault/src/impl_callbacks.rs
```diff
@@ -95,13 +96,13 @@ impl Contract {
...
-        let (i, _remaining_ctx) = match self.ctx_allocating(op_id) {
+        let (i, _, plan) = match self.ctx_allocating(op_id) {
             Ok(v) => v,
             Err(e) => return self.stop_and_exit(Some(&e)),
         };
 
-        if i != market_index {
-            return self.stop_and_exit(Some(&Error::IndexDrifted(i, market_index)));
+        if i != &market_index {
+            return self.stop_and_exit(Some(&Error::IndexDrifted(*i, market_index)));
```
peer2f00l: Is it possible to assign markets a more stable identifier?

contract/vault/src/impl_callbacks.rs
```diff
@@ -412,44 +432,216 @@ impl Contract {
...
+            ctx,
+        );
+        self.op_state = next_state;
+        self.pay_or_signal_next_withdraw()
+    }
+
+    #[private]
+    pub fn rebalance_withdraw_01_execute_withdraw_fetch_position(
+        &mut self,
+        #[callback_result] before_balance: Result<U128, PromiseError>,
```
peer2f00l, line 485: Any problem with race condition here, e.g. someone sending the account tokens during this allocation phase, so that `before_balance` is outdated?

contract/vault/src/impl_callbacks.rs
```diff
@@ -412,44 +432,216 @@ impl Contract {
...
+        market: AccountId,
+        batch_limit: Option<u32>,
+        before_principal: U128,
+    ) -> PromiseOrValue<()> {
+        if let Err(e) = self.ctx_allocating(op_id) {
+            return self.stop_and_exit(Some(&e));
+        }
+        let Ok(before_balance) = before_balance else {
+            panic_with_message("Failed to read balance")
+        };
```
peer2f00l, line 506: If this errors, does that mean we are stuck in the allocating `op_state`?

contract/vault/src/impl_callbacks.rs
```diff
@@ -412,44 +432,216 @@ impl Contract {
...
+        &mut self,
+        #[callback_result] position: Result<Option<SupplyPosition>, PromiseError>,
+        op_id: u64,
+        market: AccountId,
+        before_principal: U128,
+        before_balance: U128,
+    ) -> PromiseOrValue<()> {
+        if let Err(e) = self.ctx_allocating(op_id) {
+            return self.stop_and_exit(Some(&e));
+        }
```
peer2f00l: Rather than having checks like this at the beginning of every state-transition-effectful function, I tinkered around to see if I could find an safer API.

The downside to this approach is because it borrows `&mut op_state` exclusively, you may need to wrap it in a `RefCell` or something to make the borrow checker happy. Or, store it in its own storage key separately from the default contract state struct.

This could also be equivalently implemented using a `Guard` wrapper struct instead of the `FnOnce` callback, which would make for less nesting.

```rust

impl TryFrom<OpState> for IdleState {
    type Error = &'static str;

    fn try_from(value: OpState) -> Result<Self, Self::Error> {
        match value {
            OpState::Idle => Ok(IdleState),
            _ => Err("expecting Idle state"),
        }
    }
}

impl TryFrom<OpState> for AllocatingState {
    type Error = &'static str;

    fn try_from(value: OpState) -> Result<Self, Self::Error> {
        match value {
            OpState::Allocating(a) => Ok(a),
            _ => Err("expecting Allocating state"),
        }
    }
}

impl OpState {
    pub fn transition<F: TryFrom<OpState>, T: Into<OpState>>(
        &mut self,
        f: impl FnOnce(F) -> T,
    ) -> Result<(), &'static str> {
        let mut state = OpState::Idle;
        std::mem::swap(&mut state, self); // or .clone(), or introduce another "processing" state...
        let Ok(from) = F::try_from(state) else {
            return Err("Unexpected state transition");
        };
        *self = f(from).into();
        Ok(())
    }
}

#[test]
fn safer_transition() {
    let mut op_state = OpState::Idle;
    op_state
        .transition::<IdleState, _>(|_| AllocatingState {
            op_id: 0,
            index: 0,
            remaining: 0,
            plan: vec![],
        })
        .unwrap();
    op_state
        .transition::<AllocatingState, _>(|mut a| {
            a.remaining = 100;
            a
        })
        .unwrap();
    op_state
        .transition::<AllocatingState, _>(|_| IdleState)
        .unwrap();
    // error
    op_state
        .transition::<AllocatingState, _>(|_| IdleState)
        .unwrap();
}
```

contract/vault/src/impl_callbacks.rs
```diff
@@ -633,31 +822,50 @@ impl Contract {
         match &self.op_state {
             OpState::Allocating(_) => self.stop_and_exit_allocating(msg),
             OpState::Withdrawing(_) => self.stop_and_exit_withdrawing(msg),
-            OpState::Payout(_) => self.stop_and_exit_payout(msg),
+            OpState::Payout(_) => {
+                let reason = msg.map(|m| Reason::Other(m.to_string()));
+                return PromiseOrValue::Promise(
+                    ext_ft_core::ext(self.underlying_asset.contract_id().into())
+                        .with_static_gas(Gas::from_tgas(5))
```
peer2f00l: Named constant

contract/vault/src/impl_callbacks.rs
```diff
@@ -633,31 +822,50 @@ impl Contract {
...
         PromiseOrValue::Value(())
     }
 
-    /// Validate current op is Allocating and return (index, remaining)
-    pub(crate) fn ctx_allocating(&self, op_id: u64) -> Result<(u32, u128), Error> {
+    /// Validate current op is Allocating and return (index, remaining, plan)
+    pub(crate) fn ctx_allocating(
+        &self,
+        op_id: u64,
+    ) -> Result<(&u32, &u128, &AllocationPlan), Error> {
```
peer2f00l: Why not return an `&AllocatingState` to preserve field names?

contract/vault/src/lib.rs
```diff
@@ -291,33 +284,52 @@ impl Contract {
...
         if let Some(id) = self.peek_next_pending_withdrawal_id() {
-            let pending = self.pending_withdrawals.get(&id).unwrap_or_else(|| {
-                templar_common::panic_with_message("pending vanished unexpectedly")
-            });
+            let pending = self
+                .pending_withdrawals
+                .get(&id)
+                .unwrap_or_else(|| env::panic_str("pending vanished unexpectedly"));
+            Event::WithdrawProgress {
+                phase: "execution_started".to_string(),
```
peer2f00l: Use enum

contract/vault/src/lib.rs
```diff
@@ -326,41 +338,52 @@ impl Contract {
...
+        if ctx.index != market_index {
+            self.op_state = OpState::Withdrawing(WithdrawingState {
+                index: market_index,
+                ..ctx
+            });
+        }
         PromiseOrValue::Promise(
             ext_ft_core::ext(self.underlying_asset.contract_id().into())
                 .with_static_gas(Gas::from_tgas(5))
+                .with_unused_gas_weight(0)
```
peer2f00l: `unused_gas_weight == 0` is a bit risky, because it prevents remedying out-of-gas errors by increasing attached gas.

contract/vault/src/lib.rs
```diff
@@ -369,6 +392,119 @@ impl Contract {
...
+    ///   refunds escrowed shares to the owner, and dequeues the pending request.
+    /// - Clears withdraw state and market execution locks and returns the vault to Idle.
+    pub fn unbrick(&mut self) -> PromiseOrValue<()> {
+        Self::assert_allocator();
+
+        match self.op_state.clone() {
+            OpState::Withdrawing(s) => {
+                let id = self.next_withdraw_to_execute;
+                Event::UnbrickInvoked {
+                    phase: "withdrawing".to_string(),
```
peer2f00l: Also enum

peer2f00l (changes requested):

contract/vault/src/impl_callbacks.rs
carrion256: Yep, i rewrote this a few times between best-effort and panic early. Now i'm reading this with fresh eyes, it should definitely be best-effort

contract/vault/src/impl_callbacks.rs
carrion256: Good question. 
If anybody sends `extra` to the vault during these two receipts the `ft_balance_of` verification in `*_03`  derives (in `compute_withdraw_deltas/settle_market_principal_and_idle`):
- `principal_delta` drop reported by the market
- `inflow := after - before` 
- `creditable = min(principal_delta, inflow)`
- `extra := inflow - principal_delta`
- `market_principal := min(before, creditable)`
- `idle += inflow` 

We take the min of the two deltas, so `extra` tokens are simply free to allocate/withdraw as they are credited to idle_balance.

contract/vault/src/impl_callbacks.rs
carrion256: Could you explain further how this might be actually simpler or safer over a simple match function? 
With the suggested approach:
- we would have to reintroduce ad‑hoc guards inside each closure.
- The TryFrom approach forces ownership moves, meaning we’d start adding Rc<RefCell<…>> or clone-heavy code to compensate.

So far this seems overengineered, since we'd still need to write the `else`. The main goal of these callback checks is to be extremely defensive about any CEI boundaries. Which is as intended.

We could make some newtype which checks op_id and market_index, or a macro like `guard_state!(Allocating, op_id, Error::NotAllocating)`

contract/vault/src/lib.rs
carrion256: Actually i disagree, i think it **hides** out-of-gas errors by relying on donations from other callbacks. We should be 100% intentional on the gas we attach.

contract/vault/src/impl_callbacks.rs
carrion256: You're correct that there is a small donation attack between 02 -> 03 here. I am now updating it to trust the FT balance_of.

contract/vault/src/impl_callbacks.rs
carrion256: The index is intrinsic in the supply queue and withdraw route. It's also a part of morpho, so we'd change the API further if we rewrote the indices.

Edit: iam sdtill thinking about this, as i think i would prefer to reduce confusion on queue changes despite the risk of diverging from morpho

Edit: okay i am going to introduce a lookup table in a separate PR, since the other (already reviewed) work would need to be rebased

contract/vault/src/impl_callbacks.rs
peer2f00l: Macro could also work. The intent is to reduce the likelihood of errors such as:
- Forgetting to check that the current state is correct (e.g. Idle),
- Forgetting to update the state (e.g. during a short-circuit return),

...by making them into compile-time errors.

contract/vault/src/lib.rs
peer2f00l: Okay, I trust that this gas value was calculated based on proper testing, then.

contract/vault/src/impl_callbacks.rs
peer2f00l: Will resolve then in interest of that PR.

peer2f00l (approved):

peer2f00l:
Some issues with tests not compiling, but otherwise seems fine.

PR #276, opened 2025-10-28
Relayer performs Pyth updates on demand

peer2f00l:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/276)


carrion256:
*[`service/relayer/src/app/args.rs` line 59 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDBxksDs7wlCKSs6RQ:-OdDBxksDs7wlCKSs6RR:bacb1nx) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/args.rs#L59)):*
> ```XML
>         default_value = "25"
>     )]
>     pub refresh: Duration,
> ```

We should *always* push on demand. Negating to push it (especially on chains with minimal updates essentially undermines the smoothing & confidence.



_Code quote:_
```XML
    /// Do not push price updates to Pyth oracle if the last push was less
    /// than this long ago, even if requested.
    #[arg(
        id = "pyth-refresh-secs",
        long = "pyth-refresh-secs",
        env = "PYTH_REFRESH_SECS",
        value_parser = duration_from_secs,
        default_value = "25"
    )]
    pub refresh: Duration,
```

carrion256:
*[`service/relayer/src/app/mod.rs` line 102 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDCM1MEGYuM1_ZNTZO:-OdDCM1MEGYuM1_ZNTZP:bbk4qvz) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/mod.rs#L102)):*
> ```XML
>     #[tracing::instrument(skip(self), fields(gas = %gas))]
>     pub async fn estimate_cost_of_gas(&self, gas: near_sdk::Gas) -> Option<NearToken> {
>         const TERA: u128 = near_sdk::Gas::from_tgas(1).as_gas() as u128;
> ```

NIT: not sure what this name even means - perhaps was a constant for the sake of avoiding magic numbers, yet still magic.

carrion256:
*[`service/relayer/src/app/mod.rs` line 269 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDCprG3k8EtkSAyFCN:-OdDCprG3k8EtkSAyFCO:b-n8yd2a) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/mod.rs#L269)):*
> ```XML
>                     function_name: call.method_name.clone(),
>                 });
>             }
> ```

This forces an iterative style error since it fails fast. I'd suggest to collect all the rejected methods first, then error with an array of indices & names. Otherwise, there could be `N` calls that can be rejected, yet hidden by this one.



_Code quote:_
```XML
        for (index, call) in calls.into_iter().enumerate() {
            if !contract_data.allowed_methods.contains(&call.method_name) {
                return Err(FunctionCallRejectionReason::UnknownFunctionName {
                    index,
                    function_name: call.method_name.clone(),
                });
            }
```

carrion256:
*[`service/relayer/src/app/mod.rs` line 330 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDDQnC6nQQq9hkijMU:-OdDDQnC6nQQq9hkijMV:boecj0m) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/mod.rs#L330)):*
> ```XML
>                     account_id: receiver_id.clone(),
>                 });
>             }
> ```

Ditto here

Perhaps also create a util func



_Code quote:_
```XML
            if !accounts.allowed_contract_data.contains_key(receiver_id) {
                return Err(PayloadRejectionReason::UnknownTransactionReceiverId {
                    account_id: receiver_id.clone(),
                });
            }
```

carrion256:
*[`service/relayer/src/app/mod.rs` line 362 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDDhCT7zPvBOzVNh_5:-OdDDhCT7zPvBOzVNh_6:b-vi9zrr) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/mod.rs#L362)):*
> ```XML
>                 interacted.insert(market_data.collateral_asset.contract_id().to_owned());
>             }
>             gas += calls.iter().map(|f| f.gas).sum::<u64>();
> ```

I think you're accounting for gas even on the rejected calls here.

carrion256:
*[`service/relayer/src/client/pyth.rs` line 24 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDEBw56R8uoEsDZA85:-OdDEBw56R8uoEsDZA86:b-dxdyb5) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/client/pyth.rs#L24)):*
> ```XML
>             http: reqwest::Client::new(),
>             args,
>             last_updated: Arc::new(Mutex::new(HashMap::new())),
> ```

This is not required if the pyth client is an actor

carrion256:
*[`service/relayer/src/client/pyth.rs` line 86 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDEM29E-LO5keKhsDK:-OdDEM29E-LO5keKhsDL:b-caq8qf) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/client/pyth.rs#L86)):*
> ```XML
>         for id in price_ids {
>             request = request.query(&[("ids[]", id)]);
>         }
> ```

This is quite brittle - i'd suggest to use some types from wormhole, they're available as a crate



_Code quote:_
```XML
        for id in price_ids {
            request = request.query(&[("ids[]", id)]);
        }
```

carrion256 (commented):
> @carrion256 reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/276)* status: all files reviewed, 7 unresolved discussions (waiting on @peer2f00l)

peer2f00l (commented):
> *[Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/276)* status: all files reviewed, 7 unresolved discussions (waiting on @carrion256)

___
*[`service/relayer/src/app/args.rs` line 59 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDBxksDs7wlCKSs6RQ:-OdtGVI-BQFxuzTxjJFT:b-yt6t7v) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/args.rs#L59)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

We should *always* push on demand. Negating to push it (especially on chains with minimal updates essentially undermines the smoothing & confidence.
</blockquote></details>

I can agree with this to an extent. The intent of this feature is to make sure that if many different users are simultaneously requesting price updates for the same ID that we only update it once (since only one update is necessary to satisfy an unlimited number of simultaneous requests). So, at least, I think it would be fair to reduce the default refresh duration down to something like 3-5 seconds.
___
*[`service/relayer/src/app/mod.rs` line 102 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDCM1MEGYuM1_ZNTZO:-OdtH0Um1RRYa78YGNcb:blczzaj) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/mod.rs#L102)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

NIT: not sure what this name even means - perhaps was a constant for the sake of avoiding magic numbers, yet still magic.
</blockquote></details>

Tera as in Teragas (Tgas)
___
*[`service/relayer/src/app/mod.rs` line 269 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDCprG3k8EtkSAyFCN:-OdtJ7Qn3Uq2OP_pao1t:b-896fix) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/mod.rs#L269)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

This forces an iterative style error since it fails fast. I'd suggest to collect all the rejected methods first, then error with an array of indices & names. Otherwise, there could be `N` calls that can be rejected, yet hidden by this one.
</blockquote></details>

Done.
___
*[`service/relayer/src/app/mod.rs` line 362 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDDhCT7zPvBOzVNh_5:-OdtUVJJBCmlv3XEFfDw:b5f3ocj) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/mod.rs#L362)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

I think you're accounting for gas even on the rejected calls here.
</blockquote></details>

Shouldn't matter - in the version of the code this comment was made on errors short circuit (so it doesn't matter), or in the new version it will return the list of errors late (still won't matter).
___
*[`service/relayer/src/client/pyth.rs` line 24 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDEBw56R8uoEsDZA85:-OdtYV-h-mrLSBHKPZbq:b-od2ywz) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/client/pyth.rs#L24)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

This is not required if the pyth client is an actor
</blockquote></details>

Assuming by "actor" you mean "sends and receives information exclusively via message-passing" I'm not sure how this follows.
___
*[`service/relayer/src/client/pyth.rs` line 86 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDEM29E-LO5keKhsDK:-OdtYDWPAMJ46SibrcOa:b-xnokpf) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/client/pyth.rs#L86)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

This is quite brittle - i'd suggest to use some types from wormhole, they're available as a crate
</blockquote></details>

I don't see a crate that does this...
___
*[`service/relayer/src/client/pyth.rs` line 86 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDEbCxEFgpgkML_Vc2:-OdtYG4D6a8daGRIT3OE:bbkck4b) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/client/pyth.rs#L86)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

This is quite brittle - i'd suggest to use some types from pyth, they're available as a crate
</blockquote></details>

I was not able to find an official Pyth crate that supports this. Doesn't seem worth adding a whole dependency for a single deserialization operation.

carrion256:
*[`service/relayer/src/app/args.rs` line 59 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDBxksDs7wlCKSs6RQ:-OdxsthM4HIEFJzyvFGZ:b-vqetxh) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/args.rs#L59)):*
<details><summary><i>Previously, peer2f00l (peer2) wrote…</i></summary><blockquote>

I can agree with this to an extent. The intent of this feature is to make sure that if many different users are simultaneously requesting price updates for the same ID that we only update it once (since only one update is necessary to satisfy an unlimited number of simultaneous requests). So, at least, I think it would be fair to reduce the default refresh duration down to something like 3-5 seconds.
</blockquote></details>

Let's reduce it down to three seconds, as we are perhaps prematurely optimising for gas here on a cheap chain.

carrion256:
*[`service/relayer/src/client/pyth.rs` line 24 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDEBw56R8uoEsDZA85:-Odxt7jp-7XYpAEaxmR8:b-xnimet) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/client/pyth.rs#L24)):*
<details><summary><i>Previously, peer2f00l (peer2) wrote…</i></summary><blockquote>

Assuming by "actor" you mean "sends and receives information exclusively via message-passing" I'm not sure how this follows.
</blockquote></details>

My point that an Arc<Mutex> is an anti pattern which introduces a lot of locking at scale. It usually means that there's poor separation of concerns within asynchronous boundaries and it's better to keep state locally by using different concurrency primitives, if at all. Otherwise, i wouldn't block this pull request but eventually we have tech debt if the relayer requires some real scale (20+ tx/s) as this will eventually lock.

carrion256:
*[`service/relayer/src/client/pyth.rs` line 86 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDEbCxEFgpgkML_Vc2:-OdxyNhN31DLIE7CeYOG:b4w8rx) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/client/pyth.rs#L86)):*
<details><summary><i>Previously, peer2f00l (peer2) wrote…</i></summary><blockquote>

I was not able to find an official Pyth crate that supports this. Doesn't seem worth adding a whole dependency for a single deserialization operation.
</blockquote></details>

Here, i've found it https://github.com/pyth-network/pyth-crosschain/blob/main/apps/hermes/client/rust/src/lib.rs

The problem is not related to complexity of implementation or binary sizes, the problem is maintenance, I'd personally prefer that the core team who run the pyth protocol maintain an integration for me. This allows us to focus within our domain rather than any implementation drift. This means we can just get updates by dependabot and update the relayer when the API changes rather than wait for prices to stop updating, which opens us up to economic attacks in the middle of the night.

carrion256 (commented):
> @carrion256 reviewed 1 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/276)* status: 9 of 14 files reviewed, 5 unresolved discussions (waiting on @peer2f00l)

peer2f00l (commented):
> *[Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/276)* status: 7 of 15 files reviewed, 4 unresolved discussions (waiting on @carrion256)

___
*[`service/relayer/src/app/args.rs` line 59 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDBxksDs7wlCKSs6RQ:-OeKwuoEBYhUdlyRva9x:b-896fix) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/args.rs#L59)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

Let's reduce it down to three seconds, as we are perhaps prematurely optimising for gas here on a cheap chain.
</blockquote></details>

Done.
___
*[`service/relayer/src/client/pyth.rs` line 24 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDEBw56R8uoEsDZA85:-OeLAxdG6p4Rjvr8W8_2:b-4blfwp) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/client/pyth.rs#L24)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

My point that an Arc<Mutex> is an anti pattern which introduces a lot of locking at scale. It usually means that there's poor separation of concerns within asynchronous boundaries and it's better to keep state locally by using different concurrency primitives, if at all. Otherwise, i wouldn't block this pull request but eventually we have tech debt if the relayer requires some real scale (20+ tx/s) as this will eventually lock.
</blockquote></details>

Maybe Mutex, but I don't think that Arc does a lot of locking...?

peer2f00l (commented):
> *[Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/276)* status: 3 of 16 files reviewed, 4 unresolved discussions (waiting on @carrion256)

___
*[`service/relayer/src/app/mod.rs` line 330 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDDQnC6nQQq9hkijMU:-OeRSeJV74gSUZi42o0R:b-896fix) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/app/mod.rs#L330)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

Ditto here

Perhaps also create a util func
</blockquote></details>

Done.
___
*[`service/relayer/src/client/pyth.rs` line 86 at r1](https://reviewable.io/reviews/Templar-Protocol/contracts/276#-OdDEbCxEFgpgkML_Vc2:-OeRTrxbDD7xQUq2ARmz:b-ecwslr) ([raw file](https://github.com/Templar-Protocol/contracts/blob/7893b602fc7668da7aef56ba542b8e0afe87070f/service/relayer/src/client/pyth.rs#L86)):*
<details><summary><i>Previously, carrion256 (carrion256) wrote…</i></summary><blockquote>

Here, i've found it https://github.com/pyth-network/pyth-crosschain/blob/main/apps/hermes/client/rust/src/lib.rs

The problem is not related to complexity of implementation or binary sizes, the problem is maintenance, I'd personally prefer that the core team who run the pyth protocol maintain an integration for me. This allows us to focus within our domain rather than any implementation drift. This means we can just get updates by dependabot and update the relayer when the API changes rather than wait for prices to stop updating, which opens us up to economic attacks in the middle of the night.
</blockquote></details>

Will add as a separate task.

service/relayer/src/client/pyth.rs
```diff
@@ -1,52 +1,158 @@
...
+                tracing::debug!("Handling request: {request:?}");
+                match request {
+                    PythRequest::Update { price_ids, send } => {
+                        #[allow(clippy::unwrap_used, reason = "Sender should not drop")]
+                        send.send(pyth.update(&price_ids).await).unwrap();
+                    }
+                }
+            }
+        }
+    }
```
carrion256, line 65: It might make sense to add some nominal delay here to avoid blocking the runtime

carrion256 (approved):

service/relayer/src/client/pyth.rs
peer2f00l: Like so? https://github.com/Templar-Protocol/contracts/pull/276/commits/c91e8a6b854c42a237b3d81546bc823af95952df

PR #288, opened 2025-11-10
Fix: market deserialization error for interest and fees fields.

vitaliybezz:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/288)

### DIFF
--- a/common/src/accumulator.rs
+++ b/common/src/accumulator.rs
@@ -98,7 +98,7 @@ mod tests {
     #[test]
     fn default_accumulator() {
         let a = Accumulator::<crate::asset::BorrowAsset>::default();
-        
+
         assert_eq!(a.get_total(), 0.into());
         assert_eq!(a.get_next_snapshot_index(), 0);
     }
--- a/common/src/accumulator.rs
+++ b/common/src/accumulator.rs
@@ -13,6 +13,17 @@ pub struct Accumulator<T: AssetClass> {
     pub pending_estimate: FungibleAssetAmount<T>,
 }
 
+impl<T: AssetClass> Default for Accumulator<T> {
+    fn default() -> Self {
+        Self {
+            total: 0.into(),
+            fraction_as_u128_dividend: U128(0),
+            next_snapshot_index: 0,
+            pending_estimate: 0.into(),
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


common/src/borrow.rs
```diff
@@ -60,13 +60,17 @@ pub enum LiquidationReason {
 pub struct BorrowPosition {
     pub started_at_block_timestamp_ms: Option<U64>,
     pub collateral_asset_deposit: CollateralAssetAmount,
+    #[serde(default)]
     borrow_asset_principal: BorrowAssetAmount,
+    #[serde(default)]
```
peer2f00l: Should be an alias of `borrow_asset_fees`

common/src/borrow.rs
```diff
@@ -60,13 +60,17 @@ pub enum LiquidationReason {
 pub struct BorrowPosition {
     pub started_at_block_timestamp_ms: Option<U64>,
     pub collateral_asset_deposit: CollateralAssetAmount,
+    #[serde(default)]
```
peer2f00l: This should still deserialize the old structs properly.

common/src/accumulator.rs
```diff
@@ -13,6 +13,17 @@ pub struct Accumulator<T: AssetClass> {
...
+impl<T: AssetClass> Default for Accumulator<T> {
+    fn default() -> Self {
+        Self {
+            total: 0.into(),
+            fraction_as_u128_dividend: U128(0),
+            next_snapshot_index: 0,
+            pending_estimate: 0.into(),
+        }
+    }
+}
```
peer2f00l: Because we cannot say that `next_snapshot_index` is `0` by default (it is kind of like an analogue of "timestamp created at"), we cannot have a correct `Default` implementation of this type.

peer2f00l (changes requested):

common/src/borrow.rs
vitaliybezz: fixed

common/src/accumulator.rs
vitaliybezz: removed

peer2f00l (approved):

PR #292, opened 2025-11-14
feat: Add static account accumulations

petarvujovic98:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/292)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,15 +13,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "addr2line"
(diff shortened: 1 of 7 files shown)


petarvujovic98:
@peer2f00l The tests pass locally for me, so I believe it's the runner bug again.
Other than that, coverage is failing, do you want me to add tests to the accumulator service?

peer2f00l (approved):
Functionally, looks good. Tests would always be good though.

PR #298, opened 2025-11-21
EIP 712 signatures

peer2f00l:
Currently-deployed UA smart contract accounts will need to upgrade and migrate in order to use this version of the contract.

This process consists of two steps:

1. Deploy the new contract code to the account.
2. Execute the `migrate` function. This function is `#[private]`, meaning it can only be called when the predecessor is the current account (i.e. a reflexive function call). This function takes arguments:

    ```json
    {
      "args": {
        "chain_id": "397",
        "from_version": "v0"
      }
    }
    ```

    The `chain_id` should be 397 for NEAR mainnet, and 398 for NEAR testnet.

This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/298)


peer2f00l:
Review this first: https://github.com/Templar-Protocol/contracts/pull/312

contract/universal-account/src/impl_migrate.rs
```diff
@@ -0,0 +1,51 @@
...
+            })
+    }
+
+    pub fn get_target_state_version() -> u32 {
+        STATE_VERSION
+    }
+
+    pub fn needs_migration() -> bool {
+        Self::get_stored_state_version() < Self::get_target_state_version()
+    }
```
carrion256, line 32: gud

service/relayer/src/app/args.rs
```diff
@@ -214,6 +214,9 @@ pub struct UniversalAccount {
         value_delimiter = ','
     )]
     pub allowed_origin: Vec<String>,
+    /// Chain ID. NEAR mainnet = 397, NEAR testnet = 398.
+    #[arg(id = "ua-chain-id", long = "ua-chain-id", env = "UA_CHAIN_ID")]
```
carrion256: Default should be the same value as `RPC_URL` (testnet)

service/relayer/src/route/universal_account/create.rs
```diff
@@ -71,7 +88,15 @@ struct KeyIdMismatchError {
...
-            Self::Passkey(m) => Ok(m.payload_unchecked().payload_unchecked().key.clone().into()),
+            Self::Passkey(m) => Ok(m
+                .message
+                .0
+                .parsed
+                .payload_ref()
+                .payload_unchecked()
+                .key
+                .clone()
+                .into()),
```
carrion256: NIT: style: i would abstract away this into a deref or a helper

service/relayer/src/route/universal_account/create.rs
```diff
@@ -119,16 +144,34 @@ pub async fn create(
...
-            let exec_args = ExecuteArgs::Passkey {
+        CreateRequest::Passkey(mws) => {
+            let key_inner = mws
+                .message
+                .0
+                .parsed
+                .payload_ref()
+                .payload_unchecked()
+                .key
+                .clone();
```
carrion256: Same

universal-account/src/lib.rs
```diff
@@ -1,55 +1,202 @@
...
+        }
+    }
+
+    #[must_use]
+    pub fn auto_version(self) -> Self {
+        Self {
+            version: Some(env!("CARGO_PKG_VERSION").to_owned()),
+            ..self
+        }
+    }
```
carrion256: I think it would be simpler to set this as the default params in the builder (like Builder::new(..)/::default(), then anything else overwrites it. This way we don't forget to manually invoke `.auto()` for anything

universal-account/src/encoding/ethereum/signature.rs
```diff
@@ -0,0 +1,114 @@
+use std::str::FromStr;
+
+use alloy::signers::Signature as AlloySignature;
+use near_sdk::serde::{self, Deserialize, Serialize};
+
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
+pub struct Signature(pub AlloySignature);
```
carrion256, line 6: Signature implements serde already https://docs.rs/alloy-primitives/1.4.1/alloy_primitives/struct.Signature.html#impl-Serialize-for-Signature (and borsh

universal-account/src/encoding/ethereum/address.rs
```diff
@@ -0,0 +1,143 @@
...
+        BorshSerialize::serialize(&bytes, writer)
+    }
+}
+
+impl BorshDeserialize for Address {
+    fn deserialize_reader<R: std::io::Read>(reader: &mut R) -> std::io::Result<Self> {
+        let bytes = <[u8; 20] as BorshDeserialize>::deserialize_reader(reader)?;
+        Ok(Self(AlloyAddress::from(bytes)))
+    }
+}
```
carrion256: Address also implements borsh & serde https://docs.rs/alloy-primitives/1.4.1/alloy_primitives/struct.Address.html#impl-BorshDeserialize-for-Address

contract/universal-account/src/impl_migrate.rs
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+        Self::write_state_version(args.output_version());
+
+        Self(
+            args.migrate()
+                .unwrap_or_else(|e| templar_common::panic_with_message(&e.to_string())),
+        )
+    }
+}
```
carrion256, line 51: I really like this approach to migrations, perhaps we can make a generic version in the library over FromState -> ToState with a Migrateable trait to implement. Not for this PR, just a note

universal-account/src/authentication/eip712/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+            .recover_address_from_prehash(&prehash)
+            .map_err(CheckSignatureError::other)?;
+
+        (recovered_address == self.0 .0)
+            .then_some(())
+            .ok_or(CheckSignatureError::InvalidSignature)
+    }
+}
+
+impl<T: serde::Serialize> Message<T> {
```
carrion256, line 68: I am pleased you were able to use alloy here, to ensure we are not enacting crypto in a vacuum.

carrion256 (commented):
Looks great! I have two small changes to be made regarding manual serde/borsh implementations.

universal-account/src/encoding/ethereum/signature.rs
peer2f00l: Unfortunately `near-sdk` also requires that all types implement `schemars::JsonSchema` as well, hence the newtype.

universal-account/src/encoding/ethereum/address.rs
peer2f00l: borsh is not implemented in the version of the package we are using, which is limited due to near-sdk MaxSRV

universal-account/src/lib.rs
peer2f00l: Added builder

universal-account/src/encoding/ethereum/signature.rs
peer2f00l: Also their `serde::Serialize` implementation produces `"{\"r\":\"0xb12461f100937e17f066aee467c767e7dd0db7968c26b36e46d7abc8f4a08087\",\"s\":\"0x2d42ea30aba81b2b82d53c1a959b4da932bb6bbf83383665ef84fc62bddf2474\",\"yParity\":\"0x0\",\"v\":\"0x0\"}"` instead of `"\"0xb12461f100937e17f066aee467c767e7dd0db7968c26b36e46d7abc8f4a080872d42ea30aba81b2b82d53c1a959b4da932bb6bbf83383665ef84fc62bddf24741b\""`, which is, imo, less nice.

universal-account/src/encoding/ethereum/address.rs
carrion256: If this is a new feature - why would we not be using the latest version of alloy? You must run `cargo update -p borsh` after bringing in alloy with borsh, serde features

universal-account/src/encoding/ethereum/signature.rs
carrion256: The problem with this reasoning is we are adding a bunch of audit surface just for niceness. Instead, we have a nested string and users have to manually figure out the parity and the chainId to fully derive the signature. Existing users (who would use allloy) can't even use our custom signature format.

In general, i find it best to rely on the library writers so we can focus on our own tech. Alloy will likely be audited, and we get their work for free if they update anything. In this case, anybody who does `cargo update` would eventually see that client services will break, because theres some new signature format in their `ToString` implementation. This is just RAM in our brains, and an extra module for no reason.

peer2f00l: `near-sdk` requires the type to implement JsonSchema, so that necessitates the wrapper type. The serialization provided by alloy is actually custom, so far as I can tell, because the standard representation of an Ethereum signature is the single `0x...` string. Example:
- https://ethvigil.com/docs/eth_sign_example_code/
- https://w3c-ccg.github.io/ethereum-eip712-signature-2021-spec/#example-3
- https://ethsigning.guide/

universal-account/src/encoding/ethereum/address.rs
peer2f00l: Because we are not on the latest version of Rust ([near-sdk max supported rust version is 1.86](https://github.com/near/near-sdk-rs?tab=readme-ov-file#near-sdk))

universal-account/src/encoding/ethereum/signature.rs
carrion256: How Ethereum contracts/geth treat encoded signatures isn’t explicitly what matters here. The `ecrecover` precompile just requires `[v, r, s]` in packed format, and any data provided to a contract ultimately comes in as a hex string.

My point is that the alloy `Signature` type is ubiquitous in current EVM tooling. It already exposes all the required fields explicitly, so we don’t have to infer anything from chainId or do extra work on the frontend or client based code. If we move to a custom format, we’re just pushing that parsing/guesswork back onto client code and increasing maintenance burden.

Redesigning a new JSON format for Ethereum signatures seems like pure overhead to me, especially when the main reason we adopted alloy was to align with existing tooling and reuse its audited logic.

I do, however, agree that we need the newtype to build JsonSchema impl's. But that we should just delegate to alloy’s existing primitives. This way  we keep compatibility with the broader EVM ecosystem, and when alloy updates its signature handling, we don’t have to chase and maintain a parallel format on our side.

universal-account/src/encoding/ethereum/address.rs
carrion256: I do not fully understand the relevance to this, alloy's MSRV is 1.85

peer2f00l: This was a miscommunication about the version of `alloy` vs `alloy-primitives`. Fixed.

universal-account/src/encoding/ethereum/signature.rs
peer2f00l: Fixed.

carrion256 (approved):

PR #302, opened 2025-11-24
feat: introduce gates for transfers and pausing based on governance

carrion256:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/302)


contract/vault/src/governance.rs
```diff
@@ -1,5 +1,106 @@
...
+            }
+        }
+    }
+
+    // Common gate for NEP-141 share transfers (ft_transfer/ft_transfer_call).
+    /// Blocks transfers when globally paused or when sending to a blocked recipient
+    /// or a known market contract.
+    fn enforce_share_transfer_gates(
+        &self,
+        // TODO: decide if we want to block share xfer to markets (seems like user error anyway)
```
carrion256: @peer2f00l whats your take here?

peer2f00l: Might as well

common/src/vault.rs
```diff
@@ -100,6 +92,43 @@ pub struct VaultConfiguration {
...
     /// The number of decimals for the share token, usually would be the same as the underlying asset.
     pub decimals: NonZeroU8,
+    /// Restrictions for this market.
+    pub restrictions: Option<Restrictions>,
+}
+
+/// Restrictions that can be applied to the vault.
+///
+/// It should cover both Whitelist style functionality and Blacklist style functionality.
+/// It should also enable Pausing
```
peer2f00l, line 102: This reads more like a prompt than documentation.

common/src/vault.rs
```diff
@@ -100,6 +92,43 @@ pub struct VaultConfiguration {
...
+#[near(serializers = [borsh, json])]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Restrictions {
+    Paused,
+    BlackList(BTreeSet<AccountId>),
+    WhiteList(BTreeSet<AccountId>),
```
peer2f00l, line 108: `std` collections are not a good idea for collections that can grow beyond a handful of items, since it will cost `O(n)` gas to r/w them.

If we use `std` collections, then we should probably do some rough gas testing and then hard limit the number of elements.

contract/vault/src/governance.rs
```diff
@@ -1,5 +1,101 @@
...
+use templar_common::{panic_with_message, vault::Restrictions};
+
 use super::*;
 
+#[near(serializers = [borsh])]
+#[derive(Default)]
+pub struct Gate {
+    /// Internal flag to bypass transfer gates for trusted internal flows
+    /// (e.g. escrow/redemption settlement).
+    bypass_share_transfer_gates: bool,
```
peer2f00l, line 50: Why do we prefer this strategy (seems like it runs the risk of forgetting to reset the flag) over having public methods that enforce restrictions, but private methods that can bypass them, or something like that?

If we stick with this, consider creating a wrapper fn that enforces the "flag reset" invariant.

peer2f00l (commented):
I see a few `FIXME` comments as well

contract/vault/src/governance.rs
carrion256: This is because we have to inject into the transfer hook and it has a minimal API. The other option is to manually recreate the transfer functionality using internal_ variants. This is not ideal since it’s unnecessary.

common/src/vault.rs
carrion256: Leftover pre design comments, not a prompt.
Will do a documentation pass shortly

common/src/vault.rs
carrion256: Realistically these are perfect for blacklist/whitelist as these collections do not grow drastically. I could use a bounded vec or a smallvec to infer the bound by types, but at runtime this would be a poor design choice.

contract/vault/src/governance.rs
carrion256: On the wrapper note, this is already a wrapper? See bypass-transfer. If I was golfing I could add it into a Drop implementation - although this is more verbose way of just cleaning up in the next line.

peer2f00l (approved):

PR #304, opened 2025-11-24
feat: adbication of safety rails

carrion256:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/304)

### DIFF
--- a/contract/vault/src/lib.rs
+++ b/contract/vault/src/lib.rs
@@ -553,7 +553,10 @@ impl Contract {
         U128(self.total_supply())
     }
 
-    /// Returns the maximum additional amount that can be deposited across all markets given current caps.
+    /// Returns a best-effort estimate of the maximum additional amount that can be deposited
+    /// across all markets given current caps and the current `supply_queue`.
+    ///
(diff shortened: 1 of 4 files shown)


contract/vault/src/lib.rs
```diff
@@ -164,6 +165,8 @@ pub struct Contract {
 
     // Keeper-provided withdraw route for the current Withdrawing op
     withdraw_route: Vec<AccountId>,
+
+    abdicator: Option<Abdicator>,
```
peer2f00l: Why an `Option`? It doesn't save storage space, and the semantics of `None` should be the same as an empty `Abdicator`.

contract/vault/src/tests.rs
```diff
@@ -1240,6 +1240,37 @@ fn governance_set_is_allocator_revoke_disallows_queue_ops() {
     c.set_supply_queue(vec![m1]);
 }
 
+#[test]
+#[should_panic = "abdicated set_is_allocator"]
```
peer2f00l: rstest case for each abdicate-able function

peer2f00l (approved):

PR #306, opened 2025-11-25
feat: multiple timelocks

carrion256:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/306)

### DIFF
--- a/contract/vault/src/governance.rs
+++ b/contract/vault/src/governance.rs
@@ -52,7 +52,6 @@ impl Timelocks {
             .iter()
             .enumerate()
             .find(|(_, entry)| find_fn(&entry.value))
-            .map(|(index, entry)| (index, entry))
     }
 }
 
@@ -429,9 +428,9 @@ impl Contract {
         self.revoke_timelocks(
             |a| matches!(a, TimelockedAction::MarketRemoval { market: mkt } if mkt == &market),
         );
-        self.markets.get_mut(&market).map(|m| {
+        if let Some(m) = self.markets.get_mut(&market) {
             m.cfg.removable_at = 0;
-        });
+        }
         Event::MarketRemovalRevoked { market }.emit();
     }
 
@@ -500,7 +499,7 @@ impl Contract {
                         members.len() < 2,
                         "Invariant violation: Cannot have more than one Guardian"
                     );
-                    require!(!members.contains(&account), "Already set to this address");
+                    require!(!members.contains(account), "Already set to this address");
                     !members.is_empty()
                 })
             }
@@ -527,10 +526,6 @@ impl Contract {
 
                 if let Some(cfg) = cfg {
                     require!(
-                        cfg.removable_at == 0,
-                        "Market removal pending, cannot change cap"
-                    );
-                    require!(
                         self.governance_timelocks
                             .seek_pending_timelock(|p| matches!(
                                 p,
@@ -539,7 +534,10 @@ impl Contract {
                             .is_none(),
                         "Market removal pending, cannot change cap"
                     );
-
+                    require!(
+                        cfg.removable_at == 0,
+                        "Market removal pending, cannot change cap"
+                    );
                     require!(new_cap != &cfg.cap, "New cap is same as current");
                     new_cap > &cfg.cap
                 } else {
@@ -592,7 +590,7 @@ impl Contract {
             TimelockedAction::GuardianChange { account } => {
                 Self::with_members_of_mut(&Role::Guardian, |members| {
                     members.clear();
-                    members.insert(&account);
+                    members.insert(account);
                 });
                 Event::GuardianSet {
                     account: account.clone(),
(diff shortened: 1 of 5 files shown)


contract/vault/src/governance.rs
```diff
@@ -401,3 +316,251 @@ impl Contract {
...
+                Event::GuardianSet {
+                    account: account.clone(),
+                }
+                .emit();
+            }
+            TimelockedAction::TimelockConfigChange { new_timelock_ns } => {
+                self.governance_timelocks.guardian_ns = new_timelock_ns.0;
+                self.governance_timelocks.timelock_config_ns = new_timelock_ns.0;
+                self.governance_timelocks.market_removal_ns = new_timelock_ns.0;
+                self.governance_timelocks.cap_ns = new_timelock_ns.0;
```
peer2f00l: What is the point in having multiple configurable fields if they're always set to the same value?

contract/vault/src/governance.rs
```diff
@@ -401,3 +316,251 @@ impl Contract {
...
+        self.governance_timelocks
+            .seek_pending_timelock(find_fn)
+            .inspect(|(_, pending)| {
+                pending.verify();
+            })
+            .map(|(i, entry)| (i, entry.clone()))
+            .map(|(i, entry)| {
+                self.governance_timelocks.pending_actions.remove(i);
+                entry.value
+            })
```
peer2f00l:
```suggestion
        let (i, entry) = self.governance_timelocks.seek_pending_timelock(find_fn)?;
        entry.verify();
        let action = entry.value.clone();
        self.governance_timelocks.pending_actions.remove(i);
        Some(action)
```

contract/vault/src/governance.rs
```diff
@@ -1,4 +1,55 @@
...
+    MarketRemoval { market: AccountId },
+}
+
+#[near(serializers = [borsh])]
+pub struct Timelocks {
+    guardian_ns: TimestampNs,
+    pub timelock_config_ns: TimestampNs,
+    cap_ns: TimestampNs,
+    market_removal_ns: TimestampNs,
+    pending_actions: VecDeque<PendingValue<TimelockedAction>>,
```
peer2f00l, line 42: It seems like the semantics for this are more like a map than a queue, since in most (if not all) places we use `seek_pending_timelock` to guarantee a maximum of one pending action per type. Of course, a vec/queue is still fine, since it will be so short that O(n) lookups are basically free (and probably cheaper than hashing to a HashMap), but I think it could simplify the code a lot more if we used a struct like:

```rust
struct PendingActions {
    guardian_change: Option<AccountId>,
    timelock_config_change: Option<U64>,
    markets: Vec<(AccountId, PendingMarketAction)>, // be sure to enforce max 1 entry per market
}

enum PendingMarketAction {
    Removal,
    CapChange(U128),
}
```

This might enforce more of the semantics at the type level / at compile-time.

contract/vault/src/governance.rs
```diff
@@ -214,115 +212,37 @@ impl Contract {
     /// If the market does not exist.
     #[payable]
```
peer2f00l: Why is this payable? Not sure if I missed an `assert_one_yocto` somewhere, but if this is necessary, it should be noted in the doc comment.

contract/vault/src/governance.rs
carrion256: A map would add complexity.

- The list is tiny
- No key scheme required, 
- Storage changes for extension
- Versioning is all easily done via extending the enum.
- Seeking can have fined grained control via the predicate
- Added branching instead of utilisation of homogeneity
- Gas is not an issue on NEAR
- Code requires extension, not modification in place => fewer state migrations. 

Pending timelocks allows multiple pending actions so long as their arguments are different, this enables you to queue future changes. The queue is ordered to enable a permissionless function "accept_change", which simply executes the next free timelock.

`accept_*` simply finds the first timelock to take that is for that action. The map does not guarantee order without having a `Map<LockKind, List<Action>`. 

For example, say a guardian is going on leave for a few weeks, but governance would take 1 week to change over. You can queue both the guardian changes ahead of time, allow governance to vote and then execute them on changeover.
I think the suggested changes are more obtuse. You have to utilise the action as a key to decide the codepath from the structure at runtime, it's not really "type safe", and harder to maintain.

contract/vault/src/governance.rs
carrion256: This is the global timelock, however, i just realised i need to add the timelock lanes, hol don!

contract/vault/src/governance.rs
```diff
@@ -568,3 +490,284 @@ impl Contract {
...
+                    }
+                    mkt.cfg.removable_at = 0;
+                }
+
+                if new_cap.0 > 0 && !was_enabled {
+                    Event::MarketEnabled {
+                        market: market.clone(),
+                    }
+                    .emit();
+                }
```
peer2f00l: Duplicate condition

peer2f00l (dismissed):

peer2f00l (approved):

PR #310, opened 2025-12-01
fix: Skip running accumulations on deleted markets

petarvujovic98:
Switch to fastnear and filter out markets that have been deleted before running accumulations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/310)

### DIFF
--- a/service/accumulator/src/main.rs
+++ b/service/accumulator/src/main.rs
@@ -94,7 +94,8 @@ async fn run_service(
     args: Args,
     shutdown: impl Future<Output = ()> + Send + 'static,
 ) -> anyhow::Result<()> {
-    let client = JsonRpcClient::connect(args.network.rpc_url());
+    let api_key = std::env::var("FASTNEAR_API_KEY").ok().unwrap_or_default();
+    let client = JsonRpcClient::connect(format!("{}?apiKey={api_key}", args.network.rpc_url()));
     let signer = Arc::new(InMemorySigner::from_secret_key(
         args.signer_account.clone(),
         args.signer_key.clone(),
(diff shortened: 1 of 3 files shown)


service/accumulator/src/main.rs
```diff
@@ -94,7 +94,8 @@ async fn run_service(
     args: Args,
     shutdown: impl Future<Output = ()> + Send + 'static,
 ) -> anyhow::Result<()> {
-    let client = JsonRpcClient::connect(args.network.rpc_url());
+    let api_key = std::env::var("FASTNEAR_API_KEY").ok().unwrap_or_default();
+    let client = JsonRpcClient::connect(format!("{}?apiKey={api_key}", args.network.rpc_url()));
```
peer2f00l: More flexible might be to specify the RPC URL as a config param, that way if it requires an inline API key like this it's supported by default. That way we don't have to tightly couple the accumulator to a single RPC provider.

service/accumulator/src/rpc.rs
```diff
@@ -385,7 +418,16 @@ pub async fn list_all_deployments(
...
+    let existing = futures::stream::iter(all_markets.into_iter())
+        .filter(|market_id| {
+            let client = client.clone();
+            let market_id = market_id.clone();
+            async move { account_exists(&client, &market_id).await.unwrap_or(false) }
+        })
+        .collect::<Vec<AccountId>>()
+        .await;
+
+    Ok(existing)
```
peer2f00l, line 430: Seems like this is breaking one of the tests.

peer2f00l (changes requested):

service/accumulator/src/main.rs
petarvujovic98: Just saw these now, will update

edit: Nevermind, just saw you fixed them

peer2f00l (approved):

PR #317, opened 2025-12-04
Stellar signatures

peer2f00l:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/317)


peer2f00l:
Still need better tests

contract/universal-account/tests/universal_account.rs
```diff
@@ -49,6 +50,7 @@ enum TestSigner {
     Passkey(p256::SecretKey),
     Ed25519Raw(ed25519_dalek::SigningKey),
     Eip712(alloy::signers::local::PrivateKeySigner),
+    Sep53(ed25519_dalek::SigningKey),
```
carrion256, line 53: What makes this an sep53 signature? It’s an eddsa 


Edit: I see, it’s just stellar naming. Can we change this to stellar?

universal-account/src/authentication/ed25519/sep53.rs
```diff
@@ -0,0 +1,33 @@
...
+    }
+}
+
+impl<T> Key<Message<T>> for VerifyKey {
+    fn check_signature(
+        &self,
+        message: &super::MessageWithSignature<Message<T>>,
+    ) -> Result<(), super::CheckSignatureError> {
+        (self.0)
+            .verify(&message.message.preimage_for_signing(), &message.signature)
```
carrion256: Is this implementation actually different across signers? Seems like I have reviewed it quite a bit

universal-account/src/encoding/stellar/public_key.rs
```diff
@@ -0,0 +1,105 @@
...
+use stellar_strkey::ed25519::PublicKey as StellarPublicKey;
+
+type ByteEncoding = [u8; 32];
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[near(serializers = [borsh])]
+pub struct PublicKey(pub ByteEncoding);
+
+impl PublicKey {
+    pub fn verify(&self, message: &[u8], signature: &crate::encoding::ed25519::Signature) -> bool {
```
carrion256: Same here

carrion256 (approved):
We do have a lot of boilerplate which is just for new types passing through to existing display. I am not entirely sure this is idiomatic but not willing to block the PR. Let’s create some proper macro to derive all of this boilerplate or just use the derive_more crate

universal-account/src/authentication/ed25519/sep53.rs
peer2f00l: Yes, it is different. However, for SEP-53 and EIP-712 it does end up forwarding to `near_sdk::env::ed25519_verify`, so I have changed it to be inlined now.

universal-account/src/encoding/stellar/public_key.rs
peer2f00l: Removed

PR #319, opened 2025-12-05
feat: Add initial market-config-cli

petarvujovic98:
Adds a market-config generator CLI which produces the following UX:

from-template
from-contract
calculate-curve

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/319)


peer2f00l:
Cursor is hidden during interrupt with Ctrl+C (sigint) https://github.com/console-rs/dialoguer/issues/77#issuecomment-669986406

peer2f00l:
Lacks a way to enter + validate an NEP-245 token (Intents token) in interactive mode.

peer2f00l:
**Edit**: I see that the validation exists, just at the end of interactive mode. It didn't suggest for my test cases because I input `intents.near` as one of my assets, hoping it would allow me to add an NEP245 token ID.

<s>We should be able to retrieve asset decimals from the respective contracts.

Note that there a few different ways to retrieve these, depending on the specific contract, but we should support at least the following:

- NEP-148 FT Metadata: `contract.near->ft_metadata({}).decimals`
- NEP-245 MT Metadata: `contract.near->mt_metadata_base_by_token_id({ token_ids: ["<token_id>"] })[0].decimals`
- Intents resolution: parse out nep141 or nep245 tokens and resolve accordingly

</s>
image

market-config-cli/src/interactive/prompts.rs
```diff
@@ -0,0 +1,939 @@
...
+        Ok(builder)
+    }
+
+    fn prompt_yield_weights(&self, mut builder: ConfigBuilder) -> CliResult<ConfigBuilder> {
+        println!("\n🎯 Yield Distribution\n");
+
+        let supply_weight: u16 = prompt_until_valid(
+            || {
+                Input::with_theme(&self.theme)
+                    .with_prompt("Supplier yield weight (e.g., 9 for 90% to suppliers)")
```
peer2f00l: This is only true if the sum of weights = 10

peer2f00l:
Should support adding more than one static yield recipient

market-config-cli/src/config/validator.rs
```diff
@@ -0,0 +1,430 @@
...
+    progress.inc(1);
+
+    // Check that supply range is reasonable
+    progress.set_message("supply minimum");
+    if config.supply_range.minimum.is_zero() {
+        return Err(CliError::Validation(
+            "Supply range minimum must be greater than zero".into(),
+        ));
+    }
+    progress.inc(1);
```
peer2f00l: It is OK for the range to be 0; there is no semantic difference between the minimum being 0 and 1, since a borrow of 0 will be rejected for other reasons (it is a no-op).

market-config-cli/src/config/validator.rs
```diff
@@ -0,0 +1,430 @@
...
+            .green()
+            .apply_to("✓ Decimals validated")
+            .to_string(),
+    );
+
+    Ok(())
+}
+
+/// # Errors
+fn validate_ranges(config: &MarketConfiguration) -> CliResult {
```
peer2f00l: These checks are already implemented in the canonical `MarketConfiguration::validate`; since they are all basic mathematical operations they probably don't need to be graphically step-by-step-ified since they will complete effectively instantly.

peer2f00l (changes requested):

peer2f00l:
Math appears to be incorrect:

```text
✔ Select interest rate model · Piecewise
✔ Starting rate at 0% utilization (e.g., 0.02) · 0.02
✔ Optimal utilization ratio (e.g., 0.80 for 80%) · 0.9
✔ Rate at optimal utilization (e.g., 0.10) · 0.23
✔ Maximum rate at 100% utilization (e.g., 0.50) · 0.34
```

Produces:

```json
  "borrow_interest_rate_strategy": {
    "Piecewise": {
      "base": "0.02000000000000000000000000000000000001",
      "optimal": "0.90000000000000000000000000000000000001",
      "rate_1": "0.22999999999999999999999999999999999999",
      "rate_2": "0.34"
    }
  },
```

Which actually yields:

```text
Piecewise(Piecewise { params: PiecewiseParams { base: 0.02000000000000000000000000000000000001, optimal: 0.90000000000000000000000000000000000001, rate_1: 0.22999999999999999999999999999999999999, rate_2: 0.34 }, i_negative_rate_2_b: 0.079 })
At 0: 0.02
At 0.9: 0.227
At 1: 0.261
```

vitaliybezz (commented):
I would suggest to put the project under `tools/` directory, not keep it in the root.

peer2f00l (approved):
Looks good for now, further issues can be addressed in individual PRs.

PR #328, opened 2025-12-12
fix: Guarantee unique nonces on higher concurrency

petarvujovic98:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/328)


peer2f00l (commented):
This seems to significantly reduce the frequency of the nonce-desync errors, but they are still present:

```text
2025-12-16T10:27:58.115706Z ERROR run_borrow_accumulations{concurrency=10}:accumulate{borrow=AccountId("8185db6dc140.user0.tmplr.near") method="apply_interest"}: templar_accumulator: Accumulation failed: Failed to send transaction: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 136780705000074, ak_nonce: 136780705000077 }]
2025-12-16T10:27:58.115764Z  INFO run_borrow_accumulations{concurrency=10}:accumulate{borrow=AccountId("8100773487f7.user0.tmplr.near") method="apply_interest"}: templar_accumulator: Starting accumulation for market: iethwbtc-iethusdc.v1.tmplr.near
2025-12-16T10:27:59.164939Z ERROR run_borrow_accumulations{concurrency=10}:accumulate{borrow=AccountId("e40f1ae8b2a1.user0.tmplr.near") method="apply_interest"}: templar_accumulator: Accumulation failed: Failed to send transaction: handler error: [An error happened during transaction execution: InvalidNonce { tx_nonce: 136780705000079, ak_nonce: 136780705000081 }]
```

The way that I have addressed this in other services (e.g. the relayer) is to bound concurrency by the # of private keys that are provided: since a single account can have multiple private keys, we can give the accumulator e.g. 3 private keys, and then have a (max) concurrency of 3, each task using a different private key.

peer2f00l:
I suppose that another way to do this would be with the actor model, i.e. have a queue of transactions waiting to be signed, and sign them as nonces become available (i.e. the last transaction to use the key has round-tripped so that we know which nonce to use).

peer2f00l:
Will merge this just to get the improvement in, but it should be noted that the issue persists.

peer2f00l (approved):

PR #333, opened 2025-12-17
Vault pre audit

carrion256:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/333)


common/src/vault/mod.rs
```diff
@@ -1,26 +1,142 @@
...
+    fn from(value: MarketId) -> Self {
+        U64(value.0.into())
+    }
+}
+
+impl From<U64> for MarketId {
+    fn from(value: U64) -> Self {
+        Self(value.0 as u32)
+    }
+}
```
peer2f00l: Wondering why we need to support the string-serializing type here when it's a u32 internally, which should never require string serialization.

common/src/vault/external.rs
```diff
@@ -0,0 +1,70 @@
+use super::*;
+
+#[near_sdk::ext_contract(ext_vault)]
+pub trait VaultExt {
```
peer2f00l: Why not have the vault contract implement this trait? It seems like an unnecessary opportunity for de-synchronization in the codebase.

common/src/vault/mod.rs
```diff
@@ -404,6 +469,15 @@ impl AsRef<WithdrawingState> for OpState {
     }
 }
 
+impl AsRef<RefreshingState> for OpState {
+    fn as_ref(&self) -> &RefreshingState {
+        match self {
+            OpState::Refreshing(s) => s,
+            _ => panic!("OpState::Refreshing expected"),
```
peer2f00l: `AsRef` implementation must not panic: https://doc.rust-lang.org/std/convert/trait.AsRef.html#relation-to-borrow

common/src/vault/mod.rs
```diff
@@ -613,22 +695,30 @@ pub enum Event {
     #[event_version("1.0.0")]
     PerformanceFeeMintFailed { error: String },
     #[event_version("1.0.0")]
-    LockChange { is_locked: bool, market_index: u32 },
+    ManagementFeeAccrued { recipient: AccountId, shares: U128 },
```
peer2f00l: Consider moving `Event` enum + related to another file because it's really long

common/src/vault/params.rs
```diff
@@ -0,0 +1,62 @@
...
+
+pub const DAY_NS: u64 = 86_400_000_000_000;
+pub const YEAR_NS: u64 = 365 * DAY_NS;
+
+pub const MIN_TIMELOCK_NS: u64 = 0;
+pub const MAX_TIMELOCK_NS: u64 = 30 * DAY_NS;
+pub const MAX_QUEUE_LEN: usize = 64;
+
+// Fetching a position
+const GET_SUPPLY_POSITION: u64 = 4;
```
peer2f00l, line 21: `Gas::saturating_add`, etc. are all `const`, so the raw `u64` constants like this should be unnecessary.

common/src/vault/wad.rs
```diff
@@ -44,22 +51,40 @@ impl Number {
...
+        self.0.write_as_little_endian(&mut b32);
+        if b32[16..].iter().any(|&b| b != 0) {
+            u128::MAX
+        } else {
+            let mut b16 = [0u8; 16];
+            b16.copy_from_slice(&b32[..16]);
+            u128::from_le_bytes(b16)
+        }
+    }
+
```
peer2f00l, line 64:
```suggestion

    #[inline]
    #[must_use]
    pub fn as_u128_saturating(self) -> u128 {
        if self.0 .0[2] != 0 || self.0 .0[3] != 0 {
            u128::MAX
        } else {
            self.0.as_u128()
        }
    }

```

peer2f00l (changes requested):

common/src/vault/params.rs
carrion256: This isn't an optimisation - it's a usability feature so anybody else can use our gas calculations, and we can be far less magic about any arithmetic. This avoids the need for "optimistic" gas overflow protection that you currently do in market which relies on unused gas weight.

Edit: just before prod we can relax this, but its better not to make any assumptions about what gas we really need at this stage

contract/vault/src/governance.rs
```diff
@@ -648,50 +1097,220 @@ impl Contract {
...
+                    .emit();
+                }
+
+                if !performance_recipient_changed {
+                    self.fees.performance.recipient = fees.performance.recipient.clone();
+                }
+
+                if !management_recipient_changed {
+                    self.fees.management.recipient = fees.management.recipient.clone();
+                }
```
peer2f00l: Why are these conditions negated?

contract/vault/src/impl_callbacks.rs
```diff
@@ -773,14 +805,188 @@ impl Contract {
...
+
+    #[private]
+    #[allow(clippy::too_many_lines)]
+    pub fn resync_idle_balance_01_settle(
+        &mut self,
+        #[callback_result] balance: Result<U128, PromiseError>,
+        op_id: u64,
+        caller: AccountId,
+        before_idle: U128,
+        _started_at_ns: u64,
```
peer2f00l, line 860: If this argument is unused, why not remove it?

contract/vault/src/impl_callbacks.rs
```diff
@@ -773,14 +805,188 @@ impl Contract {
...
     }
+
+    #[private]
+    pub fn refresh_01_settle(
+        &mut self,
+        #[callback_result] position: Result<Option<SupplyPosition>, PromiseError>,
+        market_id: MarketId,
+        op_id: u64,
+        index: u32,
+        _before: U128,
```
peer2f00l: If this argument is unused, why not remove it?

contract/vault/src/lib.rs
```diff
@@ -261,13 +261,9 @@ impl Contract {
...
             last_refresh_ns: 0,
-            refresh_cooldown_ns: refresh_cooldown_ns
-                .map(|v| v.0)
-                .unwrap_or(DEFAULT_REFRESH_COOLDOWN_NS),
+            refresh_cooldown_ns: refresh_cooldown_ns.map_or(DEFAULT_REFRESH_COOLDOWN_NS, |v| v.0),
             idle_resync_last_ns: 0,
-            idle_resync_cooldown_ns: idle_resync_cooldown_ns
-                .map(|v| v.0)
-                .unwrap_or(120_000_000_000),
+            idle_resync_cooldown_ns: idle_resync_cooldown_ns.map_or(120_000_000_000, |v| v.0),
```
peer2f00l: Magic number

peer2f00l (changes requested):
Literally minuscule nitpicks. Looks fantastic.

peer2f00l (approved):

PR #334, opened 2025-12-18
fix: Liquidation full strategy and improve logging

vitaliybezz:
- **Fix Full Liquidation Strategy**: Corrects calculation to properly use 100% of available funds up to the liquidatable amount
- **Fix Asset Logging**: Replace asset symbols with asset IDs to prevent confusion when different token types share the same symbol
- **Add Token Swap Script**: Introduces `intent-swap.sh` for automated NEP-141 to NEP-245 token swaps via NEAR Intents 1-Click API
- **Implement** automatic Pyth price updates for stale oracle data

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/334)


service/liquidator/scripts/intent-swap.sh
```diff
@@ -0,0 +1,238 @@
...
+#   1. Request quote from 1-Click API (get deposit address)
+#   2. Register storage deposit for deposit address
+#   3. Transfer NEP-141 tokens to deposit address
+#   4. Submit transaction hash to 1-Click API
+#   5. Poll status until SUCCESS/REFUNDED/FAILED
+#
+# ==============================================================================
+
+# Configuration
+NEAR_ACCOUNT="tmplr-liq.near"
```
royalf00l: There should be an easy way for the liquidator to change this to their own account otherwise it could be hard to find

royalf00l: Or just make a big note to change this config

vitaliybezz: I will remove this hardcode to generic one.
This script is just to simplify swapping.

vitaliybezz: Will add note too.

service/liquidator/scripts/intent-swap.sh
```diff
@@ -0,0 +1,238 @@
...
+# Phase 2: Transfer Tokens
+# ==============================================================================
+echo "Phase 2: Transferring tokens..."
+
+# Register storage deposit (may fail if already registered - that's OK)
+echo "Registering deposit address..."
+near contract call-function as-transaction "$FROM_CONTRACT" storage_deposit \
+  json-args "{\"account_id\":\"$DEPOSIT_ADDRESS\",\"registration_only\":true}" \
+  prepaid-gas '30 Tgas' \
+  attached-deposit '0.00125 NEAR' \
```
royalf00l, line 137: In case the script fails, you may need to attach more prepaid-gas and attached-deposit. The prepaid gas gets refunded if you overpay on NEAR

royalf00l (commented):
Lmk after it is successfully tested and I'll approve

service/liquidator/scripts/intent-swap.sh
vitaliybezz: This script for manual execution to speed up onboarding. 
What are the solid average numbers, so transaction go through with high probability? We can also make them configurable and add note.

royalf00l: Let's just make these the default and add a note to increase gas if txn fails

royalf00l (approved):

PR #360, opened 2026-02-02
feat: virtual supply tracking

peer2f00l:
This PR introduces virtual supply tracking.

Active supply is split into two categories: **real** and **virtual**. Supply that is deposited directly by suppliers (this is the normal workflow) is considered **real**. Supply that is added to a supplier's record via yield compounding is **virtual** supply. The sum of active + virtual is still used to calculate information such as usage ratio, interest rates, yield, etc.

Supply must be considered **real** in order for it to be withdrawn. The mechanism to convert virtual supply to real supply is as follows:
1. Supplier Susie deposits $100 in supply. Susie's record now contains: **real**: $100, **virtual**: $0.
2. Some time passes, and a compounding yield accumulation is run on Susie's account. Now Susie's record contains: **real**: $100, **virtual**: $10.
3. Susie is earning yield on $110, but can only withdraw $100.
4. Borrower Bob repays a loan, paying $5 in fees/interest. This adds $5 to the market's virtual credit record.
5. Susie runs a normal (non-compounding) accumulation. Because there is $5 in the market's virtual credit record, $5 from Susie's record are converted from virtual to real. Susie's record is now: **real**: $105, **virtual**: $5, and the market's virtual credit record is $0.
6. Susie is earning yield on $110 and can withdraw $105.

This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/360)

### DIFF
--- a/common/src/supply.rs
+++ b/common/src/supply.rs
@@ -20,14 +20,19 @@ pub struct YieldAccumulationProof(());
 #[derive(Default, Debug, Clone, PartialEq, Eq)]
 #[near(serializers = [json, borsh])]
 pub struct Deposit {
-    pub active: BorrowAssetAmount,
+    pub active_real: BorrowAssetAmount,
+    pub active_virtual: BorrowAssetAmount,
     pub incoming: Vec<IncomingDeposit>,
     pub outgoing: BorrowAssetAmount,
 }
 
 impl Deposit {
+    pub fn active(&self) -> BorrowAssetAmount {
+        self.active_real + self.active_virtual
+    }
+
     pub fn total(&self) -> BorrowAssetAmount {
-        let mut total = self.active + self.outgoing;
+        let mut total = self.active_real + self.active_virtual + self.outgoing;
         for incoming in &self.incoming {
             total += incoming.amount_real + incoming.amount_virtual;
         }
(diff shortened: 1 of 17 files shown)


common/src/supply.rs
```diff
@@ -272,44 +281,53 @@ impl<'a> SupplyPositionGuard<'a> {
...
-                    amount,
+                    amount_real,
+                    amount_virtual,
                 });
         }
     }
 
     /// Returns the amount successfully removed from incoming.
-    fn remove_incoming(&mut self, amount: BorrowAssetAmount) -> BorrowAssetAmount {
+    fn remove_incoming_real(&mut self, amount: BorrowAssetAmount) -> BorrowAssetAmount {
```
carrion256, line 298: Potential zero‑amount incoming entries: 

`record_yield_compound` always calls `add_incoming(0, amount, idx)` even when `amount == 0`, and `add_incoming` will push a new `IncomingDeposit` if there isn’t an entry for that snapshot. 

This can create empty batches and consume one of the `MAX_INCOMING` slots. Suggestion: return when `amount.is_zero()` in
  `record_yield_compound`, or a guard in add_incoming that rejects `(amount_real == 0 && amount_virtual == 0)`. 

This also protects record_deposit if 0‑amount transfers are possible.

peer2f00l:
Closing in favor of #366

PR #364, opened 2026-02-09
Add JIT swaps retry logic, and batch collateral rebalancing to liquidator

vitaliybezz:
- USD threshold check to skip swaps below configurable minimum (MIN_SWAP_VALUE_USD)
- Swap retry with exponential backoff and error classification (transient vs permanent)
- Batch swap of accumulated collateral inventory at cycle start
- Collateral price map for USD value estimation via oracle
- Improved error typing in 1-Click swap provider (SwapErrorKind)
- Tested on alpha markets

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/364)

### DIFF
--- a/service/liquidator/src/swap/retry.rs
+++ b/service/liquidator/src/swap/retry.rs
@@ -13,30 +13,38 @@ use tokio::time::sleep;
 use crate::rpc::{AppError, AppResult};
 
 /// Classification of swap errors for retry decisions.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, thiserror::Error)]
 pub enum SwapErrorKind {
     /// Amount below bridge/swap minimum (not retryable, batchable)
+    #[error("Amount too low: {message}")]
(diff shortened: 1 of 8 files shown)


royalf00l (approved):

service/liquidator/src/swap/retry.rs
peer2f00l, line 1: Would recommend using `thiserror` to define custom error types. ([example](https://github.com/Templar-Protocol/contracts/blob/a86d4d744499628befb53760e069ef85d195dcd3/common/src/asset.rs#L371-L381)]

peer2f00l (changes requested):

service/liquidator/src/swap/retry.rs
vitaliybezz: Make sense.

peer2f00l (approved):
excellent!

PR #366, opened 2026-02-10
chore: remove yield compounding

peer2f00l:
This PR removes `HarvestYieldMode::Compounding` as an option for the yield accumulation. Compounding should instead be performed by executing a withdrawal and immediately re-depositing the withdrawn yield back into the contract.

This greatly simplifies the contract's internal accounting, and eliminates some accounting edge-cases.

Withdrawals will now withdraw from accumulated yield first, meaning that there is no extra step necessary to "move" accumulated yield into the active supply before it can be withdrawn, which should actually be an end-user experience improvement as well, at least in terms of latency.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/366)


peer2f00l:
@carrion256 Considering this approach instead of virtual supply (e.g. #360) since the edge cases & complexity w.r.t. virtual supply were adding a lot of code, so I think that this simplification is vastly superior. Unfortunately it adds a little more complexity to the FE yet again, but hopefully the simplification in terms of the actual number and complexity of actions that can be performed on the contract is better in the long-term.

carrion256 (approved):

PR #370, opened 2026-02-11
chore: every package versioned independently + bump market to 1.3.0

peer2f00l:
@carrion256 Just making sure that you're aware of this -- the vault previously was using the workspace version, but I don't think that it is correct for everything to be using the same version anymore, since the repository contains a lot of different packages now. Please update the vault version to whatever you think is appropriate.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/370)

### DIFF
--- a/contract/vault/Cargo.toml
+++ b/contract/vault/Cargo.toml
@@ -3,7 +3,7 @@ edition.workspace = true
 license.workspace = true
 name = "templar-vault-contract"
 repository.workspace = true
-version = "1.2.1"
+version = "1.0.0"
 
 [lib]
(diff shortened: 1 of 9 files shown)


contract/vault/Cargo.toml
```diff
@@ -3,7 +3,7 @@ edition.workspace = true
 license.workspace = true
 name = "templar-vault-contract"
 repository.workspace = true
-version.workspace = true
+version = "1.2.1"
```
carrion256:
```suggestion
version = "1.0.0"
```

carrion256 (approved):

PR #372, opened 2026-02-20
chore: ibtc-iethusdc v1.3.0 templar alpha market configuration

peer2f00l:
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/372)

### DIFF
--- a/contract/market/examples/config/alpha/ibtc-iethusdc.templar-alpha.near.json
+++ b/contract/market/examples/config/alpha/ibtc-iethusdc.templar-alpha.near.json
@@ -1,0 +1,50 @@
+{
+  "time_chunk_configuration": { "duration_ms": "600000" },
+  "borrow_asset": {
+    "Nep245": {
+      "contract_id": "intents.near",
+      "token_id": "nep141:eth-0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48.omft.near"
+    }
+  },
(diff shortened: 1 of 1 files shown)


peer2f00l:
One thing to confirm: do we want to keep the maximum borrow? I believe we had set this parameter on the previous version of the market because it didn't support partial liquidations, so I think we should be safe to remove it now.

contract/market/examples/config/alpha/ibtc-iethusdc.templar-alpha.near.json
```diff
@@ -0,0 +1,50 @@
...
+  "borrow_interest_rate_strategy": {
+    "Piecewise": {
+      "base": "0",
+      "optimal": "0.9",
+      "rate_1": "0.0888888888888888888888888888888888889",
+      "rate_2": "2.4"
+    }
+  },
+  "borrow_maximum_duration_ms": null,
+  "borrow_range": { "minimum": "1", "maximum": "100000000000000" },
```
royalf00l: You can remove the maximum borrow

royalf00l (approved):

peer2f00l:
`ibtc-iethusdc.templar-alpha.near`

https://nearblocks.io/txns/4Nr3WE3thabXgwTdvNWgGGuUTtGzeVhUzJZfgcjLfvrs

PR #393, opened 2026-04-01
docs: proxy oracle

peer2f00l:
Removed linkcheck because it is apparently not compatible with this latest version of mdbook.

This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/393)


royalf00l (commented):
## Review: docs: proxy oracle (PR #393)

Good documentation PR. The architecture doc and operations runbook are well-structured and will be immediately useful for operators and curators. The `SignerArgs` rename and mdBook upgrade are clean housekeeping changes.

### What's Good

- **Architecture doc** is clear and accurately describes the system. Good use of the rekt.news reference for motivation. The layered freshness controls explanation is helpful.
- **Operations runbook** is thorough — governance lifecycle, CLI examples, JSON feed definition patterns, verification checklist, and the security note about `--secret-key` leaking via shell history.
- **`Priority` aggregation docs** with concrete examples (weight 10/5/1 fallback) will save operators from misconfiguration.
- **`SignerArgs` rename** from `account_id` → `signer_id` improves clarity since the field is the *signer* identity, not always the target account.
- **mdBook upgrade** (v0.4.52 → v0.5.2) with SHA256 hash verification is good security practice.
- **`oracles.md` update** correctly replaces "There is currently no backup oracle available" with proxy oracle backup support.

---

### Issues

#### 1. Empty section header in runbook (minor)

`runbook.md` line 134 has a `## Proxy Definition Patterns` header with no content — it immediately jumps to `## Aggregation Semantics`. Either add content or remove the empty header.

#### 2. Linkcheck removal reduces CI safety

Removing `[output.linkcheck]` from `book.toml` and the `mdbook-linkcheck` CI step means broken links in documentation won't be caught. Was this intentional (e.g., mdbook-linkcheck incompatible with v0.5.2)? If so, consider noting why in the commit message or finding an alternative link checker.

#### 3. Governance section understates initialization risk

The architecture doc says:
> "proposed changes are delayed by a configurable TTL"

This is true post-initialization, but the TTL **defaults to zero** on deployment. The docs should mention that operators must set a non-zero TTL as part of the deployment process. The runbook's "Change Governance TTL" section also doesn't note that the initial TTL is zero and should be set immediately after deployment.

The `--execute-immediately` flag in the runbook (line 251) is only documented as "only if the created proposal TTL is zero" — it should include a note that this is typically only during initial setup, not normal operations.

#### 4. Risk Controls section could be more specific

The "Remaining risks" in the architecture doc list:
> - correlated failures across upstream sources
> - bad configuration of feed definitions or weights
> - liveness issues if too few sources remain fresh
> - operational dependency on the governance process for feed changes

From our prior security analysis, there are more specific risks worth documenting:
- **No circuit breaker / inter-source deviation threshold** — if one source reports a wildly different price, the aggregation still includes it
- **No price bounds / sanity checks** — there's no upper/lower limit on reported prices
- **Governance TTL can be reduced to zero** via `SetActionTtl`, permanently removing timelock protection
- **No emergency bypass** — all feed changes require waiting for TTL, even during an active attack

These are known limitations that operators should be aware of. Documenting them sets correct expectations and avoids surprises.

#### 5. `min_sources` semantic could be clearer

The runbook says `min_sources` is the "minimum source threshold" but doesn't clarify that it counts *sources that pass freshness filtering*, not total configured sources. An operator might think "I have 3 sources configured, so `min_sources: 2` means I'm safe" — but if 2 sources are stale, only 1 passes the filter and the aggregation returns no price. Worth a one-line clarification.

#### 6. Missing `MedianLow` confidence-splitting explanation in architecture doc

The architecture doc mentions "Confidence-Aware Aggregation" and says "The proxy aggregates over source confidence bounds, not just point estimates" but doesn't explain *how*. The runbook does explain it (splits into lower/upper bounds), but the architecture doc should at least summarize: "Each source price is split into `[price - confidence, price + confidence]` before the median is computed."

This is important because RedStone prices have zero confidence (effectively doubled in the median), which affects the aggregation behavior.

#### 7. Runbook doesn't document `Priority` + `min_sources` interaction

With `Priority` aggregation and `min_sources: 2`, the proxy requires 2 valid sources but only uses the highest-weight one. This means the backup source must be live even though its value isn't used — `min_sources` acts as a health check, not a quorum for aggregation. This is a subtle but important behavioral detail for operators.

---

### Code Changes — Clean

The `account_id` → `signer_id` rename across `signer_args.rs` and all consuming files is consistent and complete. The test update in `market.rs` matches. The `SignerArgs::new()` constructor usage in `clear.rs` is a nice cleanup over direct struct construction. No issues found.

---

### Summary

The documentation is solid and fills an important gap. The main suggestions are:
1. Fix the empty section header
2. Add a deployment note about zero-default TTL
3. Document known limitations more specifically (no circuit breaker, no price bounds, TTL can be zeroed)
4. Clarify `min_sources` semantics after filtering
5. Consider replacing the link checker or documenting why it was removed

---
_Generated by [Claude Code](https://claude.ai/code)_

peer2f00l:
@royalf00l linkcheck is not compatible with the latest version of mdbook. I'll create a low-priority task for investigating other broken-link checking solutions.

docs/src/contract/proxy-oracle/runbook.md
```diff
@@ -0,0 +1,385 @@
...
+### `MedianLow`
+
+- uses all valid sources that survive `max_age` and `max_clock_drift`
+- requires at least `min_sources` valid sources, or it returns no price
+- splits each source into lower and upper confidence bounds, sorts the resulting values, and returns the weighted lower median
+- has no explicit outlier filter beyond freshness and source-count filtering
+
+Use this when you want combination across live sources rather than a single preferred source.
+
+Example: if two equal-weight live sources resolve to `99` and `101`, `MedianLow` returns the lower median of the valid sorted set. With two clean sources, that gives a conservative combined result rather than always preferring one provider.
```
carrion256: MedianLow is not conservative for all market actions, only collateral valuation. For debt it's the opposite. We should be careful of single biased pricing. 

Perhaps when we return prices we should return an additional header containing things like confidence, price scalar, publish time and maybe some source data. This would allow future consumers to be smarter if needed.

docs/src/contract/proxy-oracle/runbook.md
```diff
@@ -0,0 +1,385 @@
...
+
+### `Priority`
+
+- uses only valid sources that survive `max_age` and `max_clock_drift`
+- requires at least `min_sources` valid sources, or it returns no price
+- selects the surviving value with the highest numeric `weight`
```
carrion256, line 148: We are slightly overloading weight here which confused me on our call. For Priority, weight = rank. For aggregation, weight means statistical weight. We should split these into separate names, priority for ordered failover and weight only for statistical aggregation.

docs/src/contract/proxy-oracle/index.md
```diff
@@ -0,0 +1,155 @@
...
+- conservative aggregation logic
+- transparent on-chain governance for feed updates
+
+Remaining risks:
+
+- correlated failures across upstream sources
+- bad configuration of feed definitions or weights
+- liveness issues if too few sources remain fresh
+- operational dependency on the governance process for feed changes
+- no inter-source deviation check; a bad source can still influence aggregation
```
carrion256, line 150: This is mitigated with windows. We should actually think about per-feed filters, confidence filters, step changes, asset sanity ranges.

docs/src/contract/proxy-oracle/index.md
```diff
@@ -0,0 +1,155 @@
...
+Governance actions may only be executed by the proxy oracle contract's owner.
+
+Proxy oracle governance is simple:
+
+- feed definitions are controlled through an on-chain proposal process
+- proposed changes are delayed by a configurable TTL
+- proposals can be inspected before execution
+- proposals can be executed or cancelled after review
+
+The TTL starts at zero on deployment. Operators should set a non-zero TTL during initial setup.
```
carrion256, line 116: We should disallow direct step back to zero. It can only ratchet down. But we should be aware of scope here, we don't want to reinvent governance tightly coupled to the oracle.

docs/src/contract/proxy-oracle/index.md
```diff
@@ -0,0 +1,155 @@
...
+
+Remaining risks:
+
+- correlated failures across upstream sources
+- bad configuration of feed definitions or weights
+- liveness issues if too few sources remain fresh
+- operational dependency on the governance process for feed changes
+- no inter-source deviation check; a bad source can still influence aggregation
+- no absolute price sanity bounds
+- governance can set the TTL back to zero, removing the delay for future proposals
```
carrion256, line 152: This should not happen if we eagerly compute expires_at. I did raise this in the PR.

TTL Reductions themselves should be delayed by the TTL. But this feels like we are rewriting the vault governance.

docs/src/contract/proxy-oracle/index.md
```diff
@@ -0,0 +1,155 @@
...
+- there is no emergency shortcut around the active TTL
+
```
carrion256, line 154: Old proposals can still lurk around and have their proposals executed at any arbitrary TTL change. 

Proposals should be only valid in the world they were submitted in, not brave new worlds.

docs/src/contract/proxy-oracle/runbook.md
```diff
@@ -0,0 +1,385 @@
...
+## Proxy Definition Patterns
+
+### Single-Source Feed
+
+Use for direct passthrough.
+
+```json
+{
+  "aggregator": {
+    "method": "MedianLow",
```
carrion256: Single source should not use medianlow, it should be an explicit passthrough or priority with 1 entry

docs/src/contract/proxy-oracle/index.md
```diff
@@ -0,0 +1,155 @@
...
+Each feed has its own aggregation policy:
+
+- `MedianLow`: conservative median across sources
+- `Priority`: highest-weight source wins
+
+This avoids a single global policy for all assets.
+
+### Confidence-Aware Aggregation
+
+The proxy aggregates over source confidence bounds, not just point estimates. This is more conservative than a simple passthrough or first-response model.
```
carrion256: We need more detail here, we should define the aggregated fields. If the market just uses the price scalar that's okay, other consumers need a little bit more.

docs/src/contract/proxy-oracle/index.md
```diff
@@ -0,0 +1,155 @@
+# Proxy Oracle Architecture
+
+The proxy oracle keeps a stable market-facing oracle interface while letting each feed use one or more underlying sources.
```
carrion256, line 3: So after thinking about mutability when @royalf00l mentioned this.

A market that was onboarded expecting a direct spot oracle can silently become a transformed feed, a lower-biased feed, or a backup-selected feed with different behavior. SO what should be mutable?

Source membership/Failover/Freshness = yes


Feed class? I think no. Same for aggregation families, priority/multi-sources

carrion256 (changes requested):
We should also disallow ttl to be set back to zero without ratcheting.

docs/src/contract/proxy-oracle/index.md
royalf00l: I agree that we should specifically define:
a) what fields should be mutable and 
b) what is the universe of possible new values post mutation (e.g. whitelists)

docs/src/contract/proxy-oracle/index.md
peer2f00l: Removes the delay for _future_ proposals. The TTL for each proposal is stored in the proposal.

docs/src/contract/proxy-oracle/index.md
peer2f00l: Windows are not implemented yet, so this documentation would be wrong to describe them.

docs/src/contract/proxy-oracle/runbook.md
peer2f00l: I will update the example to avoid the word "conservative."

It is not possible for the current implementation to return a nonzero confidence interval due to how it normalizes + aggregates prices. However, publish time and the specific price data are returned. Adding a way to retrieve resolution introspection is a feature request.

docs/src/contract/proxy-oracle/runbook.md
peer2f00l: https://github.com/Templar-Protocol/contracts/pull/403

docs/src/contract/proxy-oracle/index.md
peer2f00l: What should the ratchet unit be, and should it be configurable?

Also, this is a feature request, so it should probably be a separate ticket, not implemented on a docs PR.

docs/src/contract/proxy-oracle/index.md
peer2f00l: Okay, now we're backtracking on what was previously discussed. Originally the feeds were immutable, but now they are mutable only by the owner. If we want a guarantee of immutability, the correct approach given the current implementation is for the owner to abdicate ownership. Because the market configuration is immutable, and we are planning to deploy one price oracle per market, this is a reasonable workflow, I think. If the owner doesn't wish to abdicate governance over two feeds, then they can implement a nonzero TTL.

Regardless, even if we were to implement a feed class restriction, it actually is trivial to circumvent by simply pointing the proxy oracle at another proxy oracle and saying that it's Pyth. I think that it would be a waste of time and resources for us to implement this level of granularity into governance. This partial immutability is easily circumventable by a rogue owner--that's what the TTL is there for. If you don't like a change coming down the pipe, get your funds out now.

If we implement whitelisting, then one of the following must be true:

1. Templar maintains a dynamic whitelist onchain. This introduces more points of trust and failure, as well as liability on the part of Templar. Additionally, we would probably have to whitelist entire proxy definitions, or at least full entries, and not just oracle IDs, since, for example, a rogue owner could just proxy asset IDs maliciously, e.g. setting the price ID of USDT to that of BTC, or vice-versa.
2. The whitelist is immutably defined by the proxy oracle once upon deployment. This means that if a new oracle arrives, or if an old one breaks/changes address, then the proxy oracle is bricked.
3. ...unless the owner is allowed to run upgrade/migration on the contract, but if that is allowed, then we're practically back to the current trust model (no whitelist), because the new code could just ignore the whitelist.
4. ...unless the proxy oracle only allows upgrades to code that has been deployed by Templar, but now we inherit all of that liability/trust/failure again, same as case 1.

peer2f00l:
Ok, this PR is on hold until the major refactor to proxy oracles #403 is completed

PR #394, opened 2026-04-01
chore: remove testnet deployment actions

peer2f00l:
We are not really using these anymore, so it's just wasting GH Actions compute.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/394)


SmolinPavel (approved):
That's a good idea!

PR #486, opened 2026-06-26
feat(gateway): add vault methods-spec + Client ops (ENG-392)

peer2f00l:
Resolves [ENG-392](https://linear.app/templar-protocol/issue/ENG-392).

## What

First-class gateway coverage for the vault contract, mirroring market/registry, so the vault can be driven through the typed gateway `Client` and `SandboxHarness`. This unblocks the ENG-388 migration of the vault near-workspaces tests onto the harness.

- **methods-spec** (`gateway/methods-spec/src/vault.rs`): full read views + write ops, registered in `for_each_*_method!`.
- **dispatch** (`gateway/methods-dispatch/src/vault_impl.rs`): `DispatchRead`/`PlanWrite` mapping ops to vault contract calls. Write gas mirrors the `test-utils` vault controller's known-good values.
- **client** (`gateway/core/src/client/vault.rs`): bound `VaultClient` via `contract_views!`/`contract_writes!`, plus `ContractKind::Vault` detection.
- **harness** (`gateway/testing/src/sandbox.rs`): `SandboxHarness::deploy_vault`.
- **test** (`service/gateway/src/rpc/tests/vault_tests.rs`): exercises deposit → donate → resync end-to-end through the RPC server, in the same style as the other domain tests.

## Also in this PR (broader than vault)

- **Unify the gateway numeric wire type on `templar_primitives::SU128`**, removing the bespoke gateway `U128` (same string wire format). Touches the ft/mt/token/ref_finance specs, the RPC integration tests, and the harvest-static-yield tool.
- **Extract shared deposit plumbing** (storage registration + standard-agnostic `transfer_call`) into `methods-dispatch::token_ops`, consumed by both the market and vault dispatchers — it was previously duplicated between them.

## Deliberate scope decisions

- Item 4's "thin domain wrappers in `gateway/testing/src/ops.rs`" were **not** added — they would shadow the typed `Client` with a hand-maintained parallel API covering only a fragment of the surface. Vault coverage is a sibling RPC test instead.
- **Excluded views** (types live only in `contract/vault/near` and can't be sourced without a bad dependency): full `get_fee_anchor` (`FeeAccrualAnchor`) and `get_pending_governance_actions` (`PendingValue<TimelockedAction>`). `getFeeAnchorTimestamp` **is** included. `set_withdraw_queue` and the `after_*` callbacks are omitted as test-mechanics.

## Follow-up (not this PR)

- Migrate `contract/vault/near/tests/*` (~44 tests) onto the harness — the "Then" of ENG-392.
- ENG-395: a pre-existing flaky `accumulator` test surfaced while working here (unrelated to this change).

## Testing

- Full gateway service suite green (35/35), including the new vault test and the market deposit flows that share `token_ops`.
- `cargo check --workspace --tests` clean, clippy clean on touched crates, and the METHODS.md catalog drift test passes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/486)


peer2f00l:
@carrion256  This is mostly mechanical and doesn't need more than a cursory glance from you. `vault_impl.rs` is likely the only place you will find interesting code.

gateway/methods-spec/src/vault.rs
```diff
@@ -0,0 +1,538 @@
+use near_account_id::AccountId;
```
carrion256, line 1: Slightly onerous to have copied types here with just a suffix for an account. Feels like a missed oppurtunity to just newtype.

gateway/core/src/client/vault.rs
```diff
@@ -0,0 +1,243 @@
...
+            {
+                let near = self.inner.clone();
+                let contract_id = self.contract_id.clone();
+                move || async move { near.vault(contract_id).get_configuration(()).await }
+            },
+        )
+        .await
+    }
+
+    contract_views! {
```
carrion256, line 177: yeah let's update AGENTS.md in the vault so we dont forget this, since it wont be matched by ast-grep by any agent due to the DSL here.

carrion256 (commented):
Looks ok, just please update AGENTS.md in the vault so any robot doesnt screw with the boundary

gateway/methods-spec/src/vault.rs
peer2f00l: Do you mean you prefer this pattern over what's presently in the branch?

```rust
#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
pub struct WithVaultId {
    pub vault_id: AccountId,
}

#[derive(MethodSpec, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
#[method(read = "vault.getConfiguration", output = VaultConfiguration)]
pub struct GetConfiguration(pub WithVaultId);

#[derive(MethodSpec, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
#[method(read = "vault.getTotalAssets", output = SU128)]
pub struct GetTotalAssets(pub WithVaultId);

// ... maybe annotated with #[serde(transparent)]
```

gateway/methods-dispatch/src/vault_impl.rs
```diff
@@ -0,0 +1,584 @@
...
+        .await?
+        {
+            steps.push(tx_result);
+        }
+
+        steps.push(transfer_call_asset(
+            &ctx,
+            request.signer_account_id,
+            configuration.underlying_token,
+            body.vault_id,
```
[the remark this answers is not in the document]
peer2f00l: Already covered on the branch: `PlanWrite<vault::Deposit>` registers the vault on `configuration.underlying_token.into_nep141()` (via `ensure_storage_registration`) before `transfer_call_asset`, alongside the depositor registration on the share token. The fix is in the retained commits, not the discarded `d0c1ae1`.

gateway/methods-dispatch/src/vault_impl.rs
```diff
@@ -0,0 +1,584 @@
...
+            body.amount.0,
+            &DepositMsg::Supply,
+        )?);
+
+        Ok(OperationPlan { steps })
+    }
+}
```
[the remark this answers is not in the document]
peer2f00l: Already covered on the branch: the `Deposit` plan registers the vault on `configuration.underlying_token.into_nep141()` (via `ensure_storage_registration`) before `transfer_call_asset`. The fix is in the retained commits, not the discarded `d0c1ae1`.

gateway/methods-spec/src/vault.rs
```diff
@@ -0,0 +1,545 @@
...
+#[derive(MethodSpec, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[method(read = "vault.peekNextPendingWithdrawalId", output = PeekNextPendingWithdrawalIdResult)]
+pub struct PeekNextPendingWithdrawalId {
+    pub vault_id: AccountId,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct PeekNextPendingWithdrawalIdResult {
+    pub request_id: Option<u64>,
+}
```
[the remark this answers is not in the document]
peer2f00l: Done — `vault.queueTail` outputs `SU64` and `PeekNextPendingWithdrawalIdResult.request_id` is `Option<SU64>` on the branch.

gateway/methods-dispatch/src/vault_impl.rs
```diff
@@ -0,0 +1,584 @@
...
+    20
+);
+direct_write!(
+    vault::SetSupplyQueue,
+    set_supply_queue,
+    |body: vault::SetSupplyQueue| MarketsArg {
+        markets: body.markets
+    },
+    50,
+    vault::SET_SUPPLY_QUEUE_DEPOSIT
```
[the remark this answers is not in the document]
peer2f00l: The under-pay failure is fixed: the planner attaches `SET_SUPPLY_QUEUE_DEPOSIT * markets.len()`, a safe upper bound that never under-pays. Exact charging (to avoid over-paying on reorders/shrinks, which the contract does not refund) needs the current supply queue, but no exported contract view returns it (`list_markets_with_ids` is all markets, not the ordered queue). Tracking a `get_supply_queue` view as a follow-up rather than expanding this gateway PR into the contract.

gateway/methods-spec/src/vault.rs
carrion256: yeah should work

peer2f00l: Going to leave these flat for now.

The newtype only fits the ~28 specs that are just `vault_id` — the other 34 have extra fields and would need `#[serde(flatten)]`, so we'd end up with two shapes instead of one. And `vault_id` gets read/built all over dispatch, the client, and tests, so wrapping it just means `request.0.vault_id` and `WithVaultId { .. }` everywhere for no real gain. It's a plain `AccountId` with nothing to enforce.

Can revisit if it ever needs validation or similar.

peer2f00l:
@carrion256 I created (well, Claude generated) an AGENTS.md in https://github.com/Templar-Protocol/contracts/pull/486/commits/407e198760530a9e8d089cf4eae953e4309d7113. Let me know if that's what you were thinking.

carrion256 (approved):

PR #487, opened 2026-06-26
refactor(relayer): migrate onto the in-process gateway library (ENG-370)

peer2f00l:
Part of the ENG-367 gateway consumer-migration epic. Migrates the relayer off its bespoke NEAR RPC layer (`near.rs` + `cache.rs`) onto the in-process `templar-gateway-client`: every chain interaction — reads *and* writes — flows through typed gateway specs, and the gateway owns the full transaction lifecycle. The relayer service itself sheds **~810 LOC**; the branch adds the gateway operation-lifecycle store, the allowance ledger, and their tests.

## Gateway additions
- Read specs: **`chain.getBlock`** (gas price folded into the block summary), **`account.get`**, **`account.getAccessKey`** (nonce + permission scope) — all backed by `near_api` in gateway-core (no direct `near-jsonrpc-client`).
- Write spec **`tx.relaySignedDelegateAction`** (NEP-366 meta-transaction) + dispatch.
- **Per-step execution outcomes**: `StepStatus::{Succeeded, Reverted, Rejected}` carrying `ExecutionOutcome` (tokens_burnt, gas, per-receipt logs, return value). `OperationRecord` exposes `tokens_burnt()` (true signer cost: tx + receipts) and `final_outcome()`.
- `account.get` failure is now typed — **`GatewayError::AccountNotFound`** vs a generic query error.
- `gateway-client` gained a pooled multi-key signer for nonce-parallel relay throughput.

## Structural operation store (`PostgresStore`)
The operation store is **fully normalized**: instead of a `state` enum on a step row, lifecycle stage is encoded by **row-existence** across a chain of tables — `gateway_operations` (identity only) → `gateway_operation_plans` → `gateway_plan_steps` → `gateway_step_executions` → `gateway_step_results` → `gateway_step_outcomes` → `gateway_step_receipts`. A row in `results` implies it was executed; an `outcome` implies a result; and so on, so illegal intermediate states are unrepresentable rather than merely unwritten.
- **Reservations vs no-ops by structure**: an operation with no plan row is a reservation (reaped at startup); a plan row with no steps is a terminal no-op (retained so the idempotency key keeps resolving). `delete_reservation` only removes the former.
- **Identity is write-once**: the operation row (`id`, `idempotency_key`, signer, fingerprint, payload) is inserted once and never rewritten — **fixes `idempotency_key` being nulled on every step transition**, which had broken crash recovery and idempotent retries. Progress saves rewrite the step tree structurally, preserving the original signed-transaction bytes and audit timestamps.
- **Efficient recovery**: a `completed_at` marker (a pure function of the derived status) with a partial index drives `list_incomplete_operations` — startup recovery scans only in-flight operations, not full history.
- **Consistent reads**: hydration runs in a `REPEATABLE READ READ ONLY` snapshot, so the multi-table reconstruction can't observe a torn state mid-save.
- **Recoverable concurrency**: a per-operation in-process lock serialises the broom's reconcile against the live execute path (each re-reads fresh state under the lock); reconstruction rejects ambiguous rows (e.g. two simultaneously in-flight steps) rather than silently picking one.
- Tables/migrations are confined to a dedicated **`gateway` schema** (default).

## Relayer
- **Reads** (market/oracle/storage/account/chain) and **oracle updates** route through gateway specs.
- **Slim allowance accounting** — the gateway is the source of truth for status/gas/hash; the relayer keeps only allowances plus a **pending gateway-charge ledger**. The bespoke `transaction` table + status enum are gone. `pending_gateway_charge` holds one row per in-flight operation (`operation_key` PK = the gateway idempotency key, `gas_estimate`, `inner_spend`), so an account can have **multiple concurrent charges**. Flow: `lock_pending` (reserve under `FOR UPDATE` on the account, summing existing reservations so concurrent submissions can't overdraw) → gateway `execute_request` → `resolve_charge`, the single rule shared by the endpoint and the broom: **settle** (terminal — debit actual `tokens_burnt`, plus the locked deposit only if it executed), **release** (never landed — no allowance change), or **defer** (still in flight). Settle/release are idempotent via the ledger delete.
- **Broom** is a periodic, min-age-gated sweep that reconciles each ledger charge against the gateway `OperationRecord` through that same `resolve_charge` rule — it carries no resolution logic of its own. **Crash recovery** (`resume_incomplete_operations`) runs once at startup before serving, fatal-on-failure so the container restarts and retries (escape hatch: `IGNORE_STARTUP_RECOVERY_FAILURE`).
- **Transient gateway read failures are retryable, not rejections.** After the known-market allowlist passes, a failed `market::GetConfiguration` read (and the same on `/get_market_prices`) now returns `Failure` (typed via `ActionAllowanceError`/`SdaCheckError`) instead of a permanent "unknown market" `Rejected` — so a blip can't make a valid relay look invalid.
- **Relay endpoints are synchronous.** UA `execute` and registry `deploy` lock the gateway's committed 300 TGas budget, reconciled against actual spend.
- **UA-create** existence check fails closed on ambiguous reads — only a typed `AccountNotFound` means "create it".
- **Deleted `near.rs` + `cache.rs`** (nonce management is the gateway signer's; protocol-config fetch removed). Dropped the cached `MarketData` kitchen sink: `AccountData` is now just the discovered market-id set + the relay allowlist (a fail-safe security boundary); everything else resolves per-request through the (server-side-cached) gateway. Relay secret keys → `near_api::SecretKey`.

## Behaviour changes worth a look
- Relay/UA-relay respond after the operation completes (was fire-and-forget + async allowance reconciliation).
- Users are charged full receipt-level `tokens_burnt` (the old path charged only the transaction-outcome burn).
- A storage-deposit on a non-storage contract — or a reverted storage-deposit top-up — now errors loudly instead of being silently skipped / reported as success.

## Deploy note
Assumes a **greenfield gateway store** — no gateway-store-backed service is deployed yet. The lifecycle migration **refuses to run against a non-empty `gateway_operations` table** (legacy rows predate the structural tables and can't be reconstructed), and the relayer's slim-accounting migration refuses to cut over while pre-gateway in-flight charges remain. `PostgresStore::new` defaults tables to the `gateway` schema (out of `public`). Safe only because there's no existing operation data to migrate.

## Verification
- `cargo check` / `clippy --tests` clean across the relayer and gateway crates; gateway catalog regenerated (`gateway/METHODS.md`); offline `.sqlx` regenerated.
- `#[sqlx::test]` DB tests cover the ledger (multi-charge lock/settle/release/defer, idempotency, per-row broom age) and the structural store (round-trip of every lifecycle stage, reservation vs no-op, rejected-vs-reverted billing, `completed_at` tracking); CI provisions Postgres for the `--lib` test jobs.
- Sandbox integration tests (`tests/relayer.rs`, `tests/redstone.rs`) need a live node + Postgres (`./script/test.sh -p templar-relayer`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/487)


service/relayer/src/client/database.rs
```diff
@@ -121,30 +126,37 @@ impl Database {
...
+    operation_key,
+    transaction_hash
 FROM
-    account
-    JOIN transaction ON account.pending_transaction_hash = transaction.transaction_hash
+    "transaction"
 WHERE
-    pending_transaction_hash IS NOT NULL
+    "status" = 'pending'::transaction_status
+    AND transaction_hash IS NOT NULL
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `326657af`. The broom no longer depends on a recorded hash: it reconciles **every** pending row against the gateway operation store — the durable source of truth, which records the signer and the on-chain hash and survives relayer restarts — via `Client::operation_by_idempotency_key`, resolving each as `Settle` / `Release` / `InFlight`.

- `PendingTransaction` now keys solely on `operation_key` (no hash field, no `Option`); `transaction_hash` is retained as an audit column, set by `finalize`.
- A min-age floor (`MIN_PENDING_AGE`) skips rows still mid-submission so the broom never mistakes an in-flight op for an abandoned one.
- Native↔gateway hash conversion is a total 32-byte move, not a base58 round-trip.

service/relayer/src/app/mod.rs
```diff
@@ -443,73 +486,33 @@ impl App {
...
+                },
+            )
+            .await
+            .map_err(oracle::UpdateError::from)
+            .map_err(Arc::new)?;
 
-                Ok(Some(transaction_hash))
-            }
-        }
+        oracle::succeeded_tx_hash(result).map_err(Arc::new)
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `218d6f14` (gateway) + `91685b5c` (relayer). The gateway now captures each step's execution outcome (return value, per-receipt logs, gas/tokens burnt) directly on the write result — the data the RPC already hands back at submission — so no follow-up `tx.get` is needed.

`update_proxy_prices` reads the captured `return_value` (the proxy's per-price status map) and rejects when any requested price came back `Blocked`/`ResolveFailed` before the caller relays the user's action — restoring the pre-gateway reject-before-relay behavior. (The proxy tx succeeds on chain even when a price is circuit-breaker-blocked, which is why the status check is required.)

As a bonus, the same captured outcome let `execute_and_account` drop its own redundant `tx.get` for `tokens_burnt`.

gateway/core/Cargo.toml
```diff
@@ -14,7 +14,9 @@ moka.workspace = true
 near-account-id = { workspace = true, features = ["serde", "schemars"] }
 near-api.workspace = true
 near-contract-standards.workspace = true
+near-jsonrpc-client.workspace = true
```
peer2f00l: This is absolutely a decision that needs to be thoroughly justified.

gateway/core/src/client/chain.rs
```diff
@@ -1,13 +1,60 @@
...
+            }
+            None => BlockReference::Finality(Finality::Final),
+        };
+        let response = self
+            .json_rpc()?
+            .call(methods::block::RpcBlockRequest { block_reference })
+            .await
+            .map_err(|error| GatewayError::NearQuery(error.to_string()))?;
+        Ok((response.header.height, response.header.timestamp_nanosec))
+    }
```
peer2f00l: This information should be requestable using near_api; using the old libraries should not be necessary: https://docs.rs/near-api/latest/near_api/struct.Chain.html#method.block

gateway/core/src/operation.rs
```diff
@@ -319,19 +343,39 @@ impl SubmittedCurrentStep<'_> {
...
+        self.operation.current_step = Some(match outcome {
+            Some(outcome) => CurrentStep::Reverted {
+                transaction: self.transaction,
+                tx_hash,
+                outcome,
+            },
+            None => CurrentStep::Rejected {
+                transaction: self.transaction,
+                tx_hash,
+            },
```
peer2f00l: We should just have separate `reverted()` and `rejected()` functions, replacing the single `fail()` function, since each time we are calling `fail` it is with an inline-constructed `Some(...)` or `None`.

gateway/methods-dispatch/src/tx_impl.rs
```diff
@@ -80,6 +92,33 @@ impl<C: Send + 'static> PlanWrite<tx::Transfer, C> for Dispatch {
...
+        // transaction it signs and pays for, sent to the delegate's sender.
+        let signed_delegate_action =
+            near_api::types::transaction::delegate_action::SignedDelegateAction::try_from_slice(
+                &request.body.signed_delegate_action.0,
+            )
+            .map_err(|error| {
+                GatewayError::NearTransaction(format!("invalid signed delegate action: {error}"))
+            })?;
+
+        Ok(OperationPlan::single(PlannedTransaction {
```
peer2f00l: Consider whether we shouldn't provide a helper function that makes this idiom more compact. In particular, it seems like we have to specify `wait_until` everywhere we construct a `PlannedTransaction`, but it is always `ExecutedOptimistic`, so it seems like it might be more sane to simply have a helper function or builder type that sets it to that by default. Because we have a lot of `Ok(OperationPlan::single(PlannedTransaction { ...` -shaped code in this codebase, consider options that would save us the most lines of code here.

gateway/methods-spec/src/chain.rs
```diff
@@ -0,0 +1,31 @@
...
+use templar_gateway_macros::MethodSpec;
+use templar_gateway_types::{CryptoHash, NearToken};
+
+/// Fetch the current gas price (yoctoNEAR per unit of gas).
+#[derive(MethodSpec, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema, Default)]
+#[method(read = "chain.getGasPrice", output = GetGasPriceResult)]
+pub struct GetGasPrice {}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct GetGasPriceResult {
```
peer2f00l: Why not just return a `NearToken` value directly?

gateway/methods-spec/src/chain.rs
```diff
@@ -0,0 +1,31 @@
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub block_hash: Option<CryptoHash>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct GetBlockResult {
+    pub height: u64,
+    /// Block timestamp in nanoseconds since the Unix epoch.
+    pub timestamp_ns: u64,
+}
```
peer2f00l: Because a `near_api` block request returns so much more information than this (and it also includes the gas price for that block, possibly eliminating the need for a dedicated "get gas price" separate method), please consider whether this minimal pair of fields could be augmented, just as a very cheap future-proofing measure.

gateway/methods-spec/src/tx.rs
```diff
@@ -69,6 +72,15 @@ pub struct Transfer {
     pub amount: NearToken,
 }
 
+/// Relay a NEP-366 signed delegate action (meta-transaction): the signing
+/// account submits a transaction carrying the delegate action and pays its gas.
+/// `signed_delegate_action` is the borsh-encoded `SignedDelegateAction`.
+#[derive(MethodSpec, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[method(write = "tx.relayDelegateAction")]
```
peer2f00l: Consider sticking closer to the actual name of the NEAR action: `DelegateActions`.

gateway/store/migrations/20260629120000_operation_updated_at_default.up.sql
```diff
@@ -0,0 +1,12 @@
...
+-- store's inserts (which don't set it explicitly) violated the constraint. Give
+-- it the same default; the store re-inserts an operation on every save, so this
+-- tracks the last write time.
+ALTER TABLE gateway_operations
+ALTER COLUMN updated_at
+SET DEFAULT NOW();
+
+ALTER TABLE gateway_operation_steps
+ALTER COLUMN updated_at
+SET DEFAULT NOW();
```
peer2f00l: If we're also adding a new default schema/namespace, consider whether these table name prefixes `gateway_*` are actually necessary.

gateway/store/src/postgres.rs
```diff
@@ -80,21 +94,50 @@ struct OperationStepRow {
...
+
+    /// Connect with all of the store's objects confined to a specific Postgres
+    /// `schema`.
+    ///
+    /// The connection's `search_path` puts `schema` first, so the store's
+    /// unqualified DDL and queries resolve there. Pass `"public"` to use the
+    /// database's default schema (only safe when the store owns the database).
+    pub fn with_schema(database_url: &str, schema: &str) -> Result<Self, sqlx::Error> {
+        let options = PgConnectOptions::from_str(database_url)?
+            .options([("search_path", format!("{schema},public"))]);
```
peer2f00l, line 195: Do we need to do any validation of the `schema` string before interpolating it here? For example, if `schema` is an empty string or `"public"`? Note that if sqlx will fail fast (even panic) it might actually be OK -- this will be a fast-fail on startup for the service. However, even if this is the case, it should at least be documented in, like, a comment or something.

gateway/store/src/postgres.rs
```diff
@@ -80,21 +94,50 @@ struct OperationStepRow {
...
+            pool,
+            schema: schema.to_owned(),
+        })
     }
 
     pub async fn migrate(&self) -> Result<(), sqlx::migrate::MigrateError> {
+        // Create the schema before sqlx creates `_sqlx_migrations` (and the
+        // store's tables) in it via the connection search_path. `self.schema` is
+        // a trusted, caller-supplied identifier.
+        sqlx::query(&format!("CREATE SCHEMA IF NOT EXISTS \"{}\"", self.schema))
```
peer2f00l, line 207: This is definitely susceptible to injection. Not that `schema` should ever be an untrusted value, but it certainly could be (accidentally) _invalid_, so we should be more careful here. See if sqlx doesn't provide a better way to manage schemas, or at least use an injection-proof query method here.

gateway/store/src/postgres.rs
```diff
@@ -476,12 +536,17 @@ async fn insert_step_row(
...
 ) -> GatewayResult<()> {
     let actions = serde_json::to_value(&transaction.actions)?;
     let signed_transaction = signed_transaction
         .map(to_vec)
         .transpose()
         .map_err(|error| GatewayError::InvalidStoredOperation(error.to_string()))?;
+    let outcome = outcome
+        .map(serde_json::to_value)
+        .transpose()
+        .map_err(GatewayError::JsonSerialization)?;
```
peer2f00l: This seems a little brittle -- is this strictly better than storing each of the outcome fields individually, or having a separate outcome table?

gateway/store/src/postgres.rs
```diff
@@ -389,60 +432,72 @@ async fn insert_operation_steps(
...
+            Some(&step.outcome),
         )
         .await?;
     }
 
     let mut current_index = step_index(operation.succeeded_steps.len())?;
     if let Some(current_step) = &operation.current_step {
-        match current_step {
-            templar_gateway_core::CurrentStep::Prepared {
+        use templar_gateway_core::CurrentStep;
```
peer2f00l: Why not just put this at the top of the file? The fully-qualified name is being used elsewhere in this file, so that would save some bytes.

gateway/types/src/operation.rs
```diff
@@ -1,7 +1,36 @@
...
+/// Logs are grouped per receipt rather than flattened so that (a) consumers
+/// interpreting log *content* (e.g. detecting a token transfer) can attribute it
+/// to the executing contract — a transaction's receipts can run untrusted code,
+/// so a flat list is spoofable — and (b) receipt boundaries are preserved,
+/// including a receipt that executed but emitted no logs.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct ReceiptOutcome {
+    pub contract_id: near_account_id::AccountId,
+    pub logs: Vec<String>,
+}
```
peer2f00l, line 32: Seems strange to not save the success/fail outcome of a receipt in a struct called `ReceiptOutcome`. Not saying we _should_ include that field, but maybe we consider a different name. Or we could include that field, either or.

gateway/client/src/lib.rs
```diff
@@ -73,9 +73,35 @@ impl ClientBuilder {
         account_id: impl Into<ManagedAccountId>,
         secret_key: SecretKey,
     ) -> GatewayResult<Self> {
-        let signer = Signer::from_secret_key(secret_key).map_err(|error| {
-            GatewayError::NearTransaction(format!("invalid signer secret key: {error}"))
+        let signer = Signer::from_secret_key(secret_key)
+            .map_err(|error| GatewayError::InvalidSignerKey(error.to_string()))?;
+        Ok(self.signer(account_id, signer))
```
peer2f00l: Given the usage of the `signer` function here, it should probably be renamed to `with_signer`

service/relayer/src/app/mod.rs
```diff
@@ -315,51 +347,62 @@ impl App {
...
-                            }
-                        }
-                    }
-                    if market_data.oracle_kind.requires_proxy_update() {
-                        proxy
-                            .entry(market_data.oracle_id.clone())
-                            .or_default()
-                            .extend([market_data.collateral.price_id, market_data.borrow.price_id]);
-                    }
+    )> {
```
peer2f00l: Should probably give this return value a real name: tuple destructuring can be hazardous especially when two items are the same type but different semantics, as is here.

service/relayer/src/app/mod.rs
```diff
@@ -540,7 +571,7 @@ impl App {
     /// - If the receiver is not known.
     /// - If any of the function call actions are not allowed.
     #[tracing::instrument(skip(self, accounts, contract_data, calls))]
-    pub fn actions_are_allowed<'a>(
+    pub async fn actions_are_allowed<'a>(
         &self,
         accounts: &AccountData,
```
peer2f00l: We might want to consider moving away from the `AccountData` type just as we are from the `MarketData` type, if only to reduce LOC and ensure no cache staleness.

service/relayer/src/broom.rs
```diff
@@ -25,35 +40,122 @@ pub async fn start(
...
+    Settle {
+        transaction_hash: near_primitives::hash::CryptoHash,
+        tokens_burnt: near_sdk::NearToken,
+        succeeded: bool,
+    },
+    /// The operation never reached the gateway; release the pending slot.
+    Release,
+    /// Still in flight; leave it for a later sweep.
+    InFlight,
+}
```
peer2f00l: The varian names are inconsistent: `Settle` and `Release` are imperative verbs (i.e. instructions), but `InFlight` is an adjective phrase describing a state.

service/relayer/src/broom.rs
```diff
@@ -25,35 +40,122 @@ pub async fn start(
...
+    #[error("finalize error: {0}")]
+    Finalize(#[from] RecordTransactionError),
+}
+
+async fn reconcile(
+    database: &Database,
+    gateway: &GatewayClient,
+    pending: &PendingTransaction,
+) -> Result<(), ReconcileError> {
+    match resolve(gateway, pending).await? {
```
peer2f00l: Use more descriptive names for `reconcile` and `resolve`, since I think they could be easily mistaken for one another.

service/relayer/src/route/universal_account/relay.rs
```diff
@@ -236,50 +229,75 @@ pub async fn relay(
...
+        .map_err(|error| GatewayError::NearQuery(format!("invalid get_key response: {error}")))?;
+
+    Ok(view.map(|v| match v {
+        VersionedKeyParameters::V1(p) => p,
+        VersionedKeyParameters::V0(p) => PayloadExecutionParameters::builder_empty()
+            .with_key_parameters(p)
+            .verifying_contract(ua_account_id)
+            .build(),
+    }))
+}
```
peer2f00l, line 294: This logic belongs in the definition of the gateway operation.

service/relayer/tests/relayer.rs
```diff
@@ -76,6 +76,47 @@ use test_utils::*;
...
+    worker: &Worker<Sandbox>,
+    account_id: &near_workspaces::AccountId,
+    public_key: near_crypto::PublicKey,
+) -> AccessKeyInfo {
+    use near_jsonrpc_client::{methods, JsonRpcClient};
+    use near_jsonrpc_primitives::types::query::QueryResponseKind;
+    use near_primitives::{
+        types::{BlockReference, Finality},
+        views::QueryRequest,
+    };
```
peer2f00l: Prefer gateway libs (or `near_api` at worst).

service/liquidator/src/executor.rs
```diff
@@ -334,12 +334,19 @@ impl LiquidationExecutor {
...
     async fn first_failed_receipt(
         &self,
         result: &templar_gateway_types::common::WriteOperationResult,
     ) -> LiquidatorResult<Option<String>> {
         let Some(tx_hash) = result.operation.latest_tx_hash() else {
-            return Ok(None);
+            return Err(LiquidatorError::TransactionFailed(format!(
+                "Liquidation operation {} succeeded without a transaction hash; cannot inspect receipts",
+                result.operation.id.0
+            )));
```
peer2f00l: `TransactionFailed` seems to be semantically incorrect to represent the case "success without a transaction hash"

gateway/types/src/operation.rs
peer2f00l: Done in `56841277` — added a `ReceiptStatus { Succeeded, Failed }` field to `ReceiptOutcome`, captured from near_api's `receipt_failures()`. (The failure *message* still isn't captured — near_api keeps per-receipt status private — tracked in ENG-407.)

gateway/client/src/lib.rs
peer2f00l: Done in `1624841f` — renamed `signer` → `with_signer`.

service/relayer/src/app/mod.rs
peer2f00l: Done in `1624841f` — the 3-tuple is now a named `GroupedPriceUpdates { pyth, redstone, proxy }` struct (the two same-typed `PythUpdatesByOracle` fields were exactly the destructuring hazard).

service/relayer/src/app/mod.rs
peer2f00l: Investigated this against the `MarketData` removal. Dropping `AccountData` came out LOC-neutral and actually worse: the per-request snapshot is one atomic read, whereas splitting its fields into separate `RwLock`s would add double-locking and *reintroduce* the staleness risk you mention (fields drifting between reads). So I kept the struct but applied the cleanups the comment is really after, in `1624841f`: narrowed `resolve_market_ids` / `derive_sda_interactions` / `actions_are_allowed` to take `&HashSet<AccountId>` (they only need `market_ids`), and dropped a wasteful full-struct `.clone()` in the price-update route. Happy to revisit removal if you still want it gone.

service/relayer/src/broom.rs
peer2f00l: Done in `1624841f` — `InFlight` → `Defer` (a verb, consistent with `Settle` / `Release`).

service/relayer/src/broom.rs
peer2f00l: Done in `1624841f` — `resolve` → `classify` (determines the resolution) and `reconcile` → `reconcile_pending` (applies it).

service/relayer/src/route/universal_account/relay.rs
peer2f00l: Done in `d8ebbd62` — moved the versioned-key handling into `UniversalAccountClient::get_key`, which now normalizes the contract's V0 (`KeyParameters`) / V1 (`PayloadExecutionParameters`) response; the relayer just calls `ua::GetKey`. This also fixed a latent bug: the op previously deserialized only V1, so a direct consumer would have silently dropped legacy (V0) accounts.

service/relayer/tests/relayer.rs
peer2f00l: Not addressed yet — flagging honestly. This is a test-only helper (`fetch_access_key_info`) still using `near_jsonrpc_client` for a `ViewAccessKey` query, which no gateway spec covers. It wasn't part of the migration scope; I can migrate it to `near_api` as a quick follow-up if you'd like.

gateway/core/Cargo.toml
peer2f00l: Addressed in `6919fcc3`: `near-jsonrpc-client` (and `near-primitives`) are removed from gateway-core entirely — chain reads now go through `near_api::Chain::block()`. So the dependency is gone rather than justified.

gateway/core/src/client/chain.rs
peer2f00l: Done in `6919fcc3` — the chain client is rewritten on `near_api::Chain::block(...)` per the linked API; the old libraries are dropped.

gateway/core/src/operation.rs
peer2f00l: Done in `1624841f` — `fail(Option<outcome>)` is split into `reverted(outcome)` and `rejected()`; the driver calls the right one directly instead of constructing an inline `Some`/`None`.

gateway/methods-dispatch/src/tx_impl.rs
peer2f00l: Done in `1624841f` — added `OperationPlan::execute(signer, receiver, actions)`, which defaults `wait_until` to `ExecutedOptimistic`. Collapsed all six `OperationPlan::single(PlannedTransaction { … })` sites onto it.

gateway/methods-spec/src/chain.rs
peer2f00l: Done in `6919fcc3` — `GetBlockResult.gas_price` is a `NearToken` directly now (the block header carries it), and the dedicated gas-price method is removed.

gateway/methods-spec/src/chain.rs
peer2f00l: Done in `6919fcc3` — `GetBlock` now returns `{ height, timestamp_ns, gas_price, hash }` from the near_api block, and the separate `chain.getGasPrice` method was deleted (gas price folded in), exactly as you anticipated.

gateway/methods-spec/src/tx.rs
peer2f00l: Done in `0b958e5a` — renamed to `RelaySignedDelegateAction` / `tx.relaySignedDelegateAction`. We went with NEAR's exact type name `SignedDelegateAction` (that's what's relayed); NEAR has no `DelegateActions` type — its body type is `DelegateAction` and the action variant is `Delegate`.

gateway/store/migrations/20260629120000_operation_updated_at_default.up.sql
peer2f00l: Kept intentionally. The store can run in the database's `public` schema via `with_schema("public")` (the own-database mode), where the `gateway_*` prefix still disambiguates its tables from a co-tenant's. Inside a dedicated schema it's redundant but harmless, and keeping it makes both modes consistent. Easy to drop later if we decide to only ever run in a dedicated schema.

gateway/store/src/postgres.rs
peer2f00l: Done in `f42bf687` — `with_schema` now runs `validate_schema_identifier` (rejects empty / non-identifier / >63 bytes; `"public"` passes) and returns a `Configuration` error on bad input. Behavior is documented on the fn.

gateway/store/src/postgres.rs
peer2f00l: Done in `f42bf687` — the schema is validated as a plain `[A-Za-z_][A-Za-z0-9_]*` identifier before any interpolation, so the `CREATE SCHEMA` / `search_path` strings can't carry injection. (sqlx can't parameterize DDL identifiers, so validate-then-quote is the available fix.) Documented inline.

gateway/store/src/postgres.rs
peer2f00l: Agreed it was brittle — went fully relational in `56841277`: scalar columns (`outcome_tokens_burnt` / `outcome_total_gas_burnt` / `outcome_return_value`) on `gateway_operation_steps`, plus a `gateway_operation_step_receipts` child table, replacing the jsonb blob. The DB now enforces the shape and evolution is plain ALTER TABLE.

gateway/store/src/postgres.rs
peer2f00l: Done in `1624841f` — hoisted `CurrentStep` to the top-level import and dropped the in-fn `use`.

gateway/core/src/operation.rs
```diff
@@ -319,17 +361,39 @@ impl SubmittedCurrentStep<'_> {
         &self.transaction
     }
 
-    pub async fn succeed(self, tx_hash: CryptoHash) -> GatewayResult<()> {
+    pub async fn succeed(
```
peer2f00l: We should rename these functions (`succeed`, `reverted`, `rejected`) to use consistent part of speech ("succeed" is a verb in present tense, the others are not). We should also probably indicate within the name that this is an operation that affects the store, so probably something like `mark_succeeded` or `mark_success`.

gateway/core/src/operation_driver.rs
```diff
@@ -353,6 +362,49 @@ impl OperationDriver {
     }
 }
 
+/// Build an [`ExecutionOutcome`] from a transaction's final result — the data
+/// the RPC already returned on submission, so it needn't be re-fetched.
+fn execution_outcome(result: ExecutionFinalResult) -> ExecutionOutcome {
```
peer2f00l: It seems like this helper should be defined canonically somewhere. Would this not be better implemented as a `impl From<ExecutionFinalResult> for ExecutionOutcome`?

gateway/methods-spec/src/chain.rs
```diff
@@ -0,0 +1,26 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct GetBlockResult {
+    pub height: u64,
+    /// Block timestamp in nanoseconds since the Unix epoch.
+    pub timestamp_ns: u64,
+    /// Gas price (yoctoNEAR per unit of gas) at this block.
+    pub gas_price: NearToken,
+    pub hash: CryptoHash,
+}
```
peer2f00l: This is effectively a duplicate of `BlockSummary`. Is there a way to unify?

gateway/types/src/operation.rs
```diff
@@ -1,7 +1,50 @@
...
+
+/// The result of executing an operation's transaction on chain, captured from
+/// the submission outcome the RPC already returns — so callers get the return
+/// value, logs, and cost without a follow-up `tx.get`.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub struct ExecutionOutcome {
+    /// Total NEAR burnt across the transaction and all its receipts — the true
+    /// cost the signer paid.
+    pub tokens_burnt: NearToken,
+    pub total_gas_burnt: NearGas,
```
peer2f00l, line 42: Is it preferable to store these computed fields here, or can we store `tokens_burnt` and `gas_burnt` on the `ReceiptOutcome` and simply provide methods on `ExecutionOutcome` that sums them?

gateway/types/src/operation.rs
```diff
@@ -18,10 +61,29 @@ pub enum OperationStatus {
...
+    /// actually ran (logs, gas burnt, return value).
+    Reverted {
+        tx_hash: CryptoHash,
+        outcome: ExecutionOutcome,
+    },
+    /// Failed before a recorded on-chain execution (e.g. a submission error), so
+    /// no state changed as far as the gateway observed.
+    Rejected {
+        tx_hash: CryptoHash,
+    },
```
peer2f00l, line 130: Is it worth saving the rejection reason here? So we can return a clear, idempotent error message as well.

service/relayer/src/app/mod.rs
```diff
@@ -115,23 +123,36 @@ impl App {
...
+    ///
+    /// This sizes the allowance lock before submitting; the lock is reconciled
+    /// against the actual `tokens_burnt` afterwards, so a coarse estimate only
+    /// affects the pre-flight affordability gate, never the amount charged.
     #[tracing::instrument(skip(self), fields(gas = %gas))]
-    pub async fn estimate_cost_of_gas(&self, gas: near_sdk::Gas) -> Option<NearToken> {
-        const TERA: u128 = near_sdk::Gas::from_tgas(1).as_gas() as u128;
+    pub async fn estimate_cost_of_gas(&self, gas: Gas) -> Option<NearToken> {
+        // The latest block's header carries the gas price, so a dedicated
+        // gas-price read isn't needed.
```
peer2f00l: Fluff comment, or at least irrelevant here.

service/relayer/src/app/mod.rs
```diff
@@ -1224,24 +1138,88 @@ mod tests {
...
+pub fn to_gateway_hash(hash: &near_primitives::hash::CryptoHash) -> CryptoHash {
+    CryptoHash::from(near_api::CryptoHash(hash.0))
+}
+
+/// Convert a gateway hash into the native transaction-hash type. Both are
+/// 32-byte hashes, so this is a total byte-for-byte move.
+#[must_use]
+pub fn from_gateway_hash(hash: &CryptoHash) -> near_primitives::hash::CryptoHash {
+    near_primitives::hash::CryptoHash(hash.0 .0)
+}
```
peer2f00l: Shouldn't these also be `From<T>` implementations?

service/relayer/tests/relayer.rs
peer2f00l: Now properly addressed in `5cc82cb2` — rather than swap one raw client for another, added a first-class `account.getAccessKey` gateway read (mirrors NEAR `view_access_key`: returns nonce + permission scope). The test helper now uses the gateway (`account.getAccessKey` for the nonce + `chain.getBlock` for the block reference), so it goes through the same typed specs as the relayer. Thanks for the nudge to do it right.

service/liquidator/src/executor.rs
peer2f00l: Fixed in `905c9890` — now a dedicated `LiquidatorError::MissingTransactionHash` instead of the misleading `TransactionFailed`.

gateway/core/src/operation.rs
peer2f00l: Fixed in `905c9890` — renamed to `mark_succeeded`/`mark_reverted`/`mark_rejected` (consistent mood + signals the store mutation).

gateway/core/src/operation_driver.rs
peer2f00l: Done in `905c9890` — it's now `impl From<ExecutionFinalResult> for ExecutionOutcome` in gateway-types (near-api-types is already a dep there), and the driver's free fn is gone.

gateway/methods-spec/src/chain.rs
peer2f00l: Done in `c96b6dbd` — unified into a single `BlockSummary` in gateway-types, used as both the `chain.getBlock` result and the client return; dispatch is now a passthrough. (A `From` wasn't possible — both originals are foreign to the dispatch crate, so the orphan rule forces relocation.)

gateway/types/src/operation.rs
peer2f00l: Decision: keeping the computed totals. No consumer needs per-receipt cost today, and per-receipt storage means schema churn; revisit if that changes.

gateway/types/src/operation.rs
peer2f00l: Good idea, and it's *not* externally blocked (the submission/rejection error is available when we mark the step `Rejected`). Noted on ENG-407 as a distinct, shippable follow-up (separate from the per-receipt execution errors that issue is blocked on).

service/relayer/src/app/mod.rs
peer2f00l: Removed in `905c9890`.

service/relayer/src/app/mod.rs
peer2f00l: Can't, unfortunately — orphan rule: both `near_primitives::CryptoHash` and the gateway `CryptoHash` are foreign to the relayer crate, and gateway-types would have to take a `near_primitives` dep to host the impl, which we don't want. Keeping the free fns.

gateway/methods-spec/src/tx.rs
```diff
@@ -75,6 +78,15 @@ pub struct Transfer {
     pub amount: NearToken,
 }
 
+/// Relay a NEP-366 signed delegate action (meta-transaction): the signing
+/// account submits a transaction carrying the delegate action and pays its gas.
+/// `signed_delegate_action` is the borsh-encoded `SignedDelegateAction`.
+#[derive(MethodSpec, Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+#[method(write = "tx.relaySignedDelegateAction")]
+pub struct RelaySignedDelegateAction {
+    pub signed_delegate_action: Base64Bytes,
```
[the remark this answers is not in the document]
peer2f00l: Done in `e8aa97d9` — `signed_delegate_action` is now a validated `SignedDelegateActionInput` newtype that borsh-decodes the NEP-366 payload at the spec boundary (custom `Deserialize`); the dispatch uses `into_inner()`, no re-validation.

gateway/store/migrations/20260629130000_operation_outcome.up.sql
```diff
@@ -0,0 +1,32 @@
...
+    contract_id text NOT NULL,
+    status text NOT NULL,
+    logs text[] NOT NULL,
+    created_at timestamptz NOT NULL DEFAULT NOW(),
+    PRIMARY KEY (operation_id, step_index, receipt_index),
+    FOREIGN KEY (operation_id, step_index)
+        REFERENCES gateway_operation_steps (operation_id, step_index) ON DELETE CASCADE,
+    CONSTRAINT gateway_operation_step_receipts_status_check
+        CHECK (status IN ('succeeded', 'failed'))
+);
```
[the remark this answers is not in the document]
peer2f00l: N/A — greenfield, nothing is deployed, so there are no legacy `gateway_operation_steps` rows to backfill (same reason that migration was rewritten rather than appended to). The loader's strict semantics are fine on a fresh schema.

gateway/types/src/operation.rs
```diff
@@ -75,4 +151,24 @@ impl OperationRecord {
...
+    }
+
+    /// The execution outcome of the operation's last executed step, if any.
+    #[must_use]
+    pub fn final_outcome(&self) -> Option<&ExecutionOutcome> {
+        self.steps
+            .iter()
+            .rev()
+            .find_map(|step| step.status.outcome())
+    }
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `905c9890` — `final_outcome` now selects by `step.index` (max), matching `latest_tx_hash`, instead of relying on vector order.

service/relayer/.sqlx/query-f982dbea9d6a6fade6d9fda5e12efe74f6c28d4f37a14fd6958372ff086dd0ae.json
```diff
@@ -0,0 +1,25 @@
...
+        "Varchar",
+        "Numeric",
+        "Numeric"
+      ]
+    },
+    "nullable": [
+      true
+    ]
+  },
+  "hash": "f982dbea9d6a6fade6d9fda5e12efe74f6c28d4f37a14fd6958372ff086dd0ae"
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `3f599516` — `set_pending_transaction` now claims the slot first (`WITH claimed AS (UPDATE … WHERE pending_operation_key IS NULL)`) and inserts only `FROM claimed`, so a contended account no longer leaks an orphaned `pending` row.

service/relayer/migrations/20260626120000_relay_operation_key.down.sql
```diff
@@ -0,0 +1,33 @@
...
+SET pending_transaction_hash = t.transaction_hash
+FROM "transaction" t
+WHERE a.pending_operation_key = t.operation_key;
+
+ALTER TABLE account
+DROP COLUMN IF EXISTS pending_operation_key;
+
+-- Rows without a transaction hash cannot exist under the old schema.
+DELETE FROM "transaction"
+WHERE transaction_hash IS NULL;
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `3f599516` — the downgrade now `RAISE`s if any hashless `transaction` rows exist, instead of silently deleting them (losing pending accounting state).

service/relayer/migrations/20260626120000_relay_operation_key.down.sql
```diff
@@ -0,0 +1,33 @@
...
+ALTER COLUMN transaction_hash SET NOT NULL;
+
+ALTER TABLE "transaction"
+ADD CONSTRAINT pk__transaction PRIMARY KEY (transaction_hash);
+
+ALTER TABLE "transaction"
+DROP COLUMN IF EXISTS operation_key;
+
+ALTER TABLE account
+ADD CONSTRAINT fk__account__transaction FOREIGN KEY (pending_transaction_hash) REFERENCES "transaction" (transaction_hash);
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `3f599516` — FK added `NOT VALID` then `VALIDATE`d separately on the downgrade.

service/relayer/migrations/20260626120000_relay_operation_key.up.sql
```diff
@@ -0,0 +1,49 @@
...
+DROP CONSTRAINT IF EXISTS pk__transaction;
+
+ALTER TABLE "transaction"
+ADD CONSTRAINT pk__transaction PRIMARY KEY (operation_key);
+
+ALTER TABLE "transaction"
+ALTER COLUMN transaction_hash DROP NOT NULL;
+
+ALTER TABLE account
+ADD CONSTRAINT fk__account__transaction FOREIGN KEY (pending_operation_key) REFERENCES "transaction" (operation_key);
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `3f599516` — FK added `NOT VALID` then `VALIDATE`d separately.

service/relayer/src/app/mod.rs
```diff
@@ -48,65 +49,72 @@ pub use args::Configuration;
...
+                .secret_keys(args.ua.account_id.clone(), args.ua.secret_key.clone())
+                .await?;
+        }
+        let gateway = builder.build()?;
+        let relay_gateway = gateway
+            .clone()
+            .into_signing(args.relay.account_id.clone())?;
 
         #[allow(clippy::unwrap_used)]
         let database = Database::new(&args.database_url, kill.clone()).unwrap();
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `905c9890` — `App::new` propagates the DB-init error via `?` now instead of `unwrap()`.

service/relayer/src/broom.rs
```diff
@@ -25,35 +40,122 @@ pub async fn start(
...
+
+    let Some(operation) = gateway.operation_by_idempotency_key(&key).await? else {
+        // The operation never reached the gateway; the lock can be released.
+        return Ok(Resolution::Release);
+    };
+
+    let Some(tx_hash) = operation.latest_tx_hash() else {
+        // Submitted but no on-chain hash yet; reconcile on a later sweep.
+        return Ok(Resolution::Defer);
+    };
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `3f599516` — broom now `Release`s an operation that reached terminal failure with no tx hash (rejected pre-submission), instead of `Defer`ring it forever.

service/relayer/src/client/oracle/spec/pyth.rs
```diff
@@ -98,20 +97,29 @@ impl Spec for PythSpec {
...
+        let vaa = self
+            .latest_vaa(feed_ids)
+            .await
+            .map_err(|e| UpdateError::Fetch(Box::new(e)))?;
+
+        let result = gateway
+            .execute(pyth_spec::UpdatePriceFeeds {
+                oracle_id,
+                data: Base64Bytes(vaa),
+            })
```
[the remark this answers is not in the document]
peer2f00l: Resolved in `7667672b` — went with removing the dead config (the Pyth deposit is the contract's update fee and the gas is just the per-tx max, so these aren't good operator knobs). The gateway keeps its sensible hardcoded values; dynamic Pyth fee (`get_update_fee`) is tracked separately.

service/relayer/src/client/oracle/spec/redstone.rs
```diff
@@ -59,25 +54,31 @@ impl Spec for RedStoneSpec {
...
-            gas: near_primitives::gas::Gas::from_gas(self.config.update_gas.as_gas()),
-            deposit: self.config.update_deposit,
-        }
-        .into()])
+        let result = gateway
+            .execute(redstone_spec::WritePrices {
+                oracle_id,
+                feed_ids: feed_ids.to_vec(),
+                payload: Base64Bytes(payload),
+            })
```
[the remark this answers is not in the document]
peer2f00l: Resolved in `7667672b` — removed the dead `update_gas`/`update_deposit` config (it was wired nowhere). RedStone keeps the gateway's hardcoded gas; no deposit needed.

service/relayer/src/route/get_market_prices.rs
```diff
@@ -41,5 +59,14 @@ pub async fn get_market_prices(
...
+    let mut prices: HashMap<_, _> = result
+        .prices
+        .into_iter()
+        .map(|entry| (entry.price_id, entry.price))
+        .collect();
+
+    SimpleResponse::success(ViewMarketPrices {
+        borrow: prices.remove(&cfg.borrow_asset_price_id).flatten(),
+        collateral: prices.remove(&cfg.collateral_asset_price_id).flatten(),
+    })
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `905c9890` — uses non-consuming `get().cloned()` lookups now, so a price ID shared by borrow+collateral resolves for both.

service/relayer/src/route/universal_account/create.rs
```diff
@@ -334,34 +321,33 @@ pub async fn create(
...
         .await
     {
-        Ok(r) => r,
+        Ok(transaction_hash) => transaction_hash,
         Err(e) => {
-            tracing::error!("Failed to send account contract deployment transaction: {e}");
+            tracing::error!("Failed to deploy universal account: {e}");
             return SimpleResponse::Failure {
-                error: "Failed to send account contract deployment transaction".to_string(),
+                error: "Failed to deploy universal account".to_string(),
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `e8aa97d9` — `execute_and_account` now returns `{ transaction_hash, succeeded }` and the UA create route returns failure when the deploy reverted (accounting is still finalized).

service/relayer/src/route/universal_account/relay.rs
```diff
@@ -136,15 +135,20 @@ pub async fn relay(
...
-                .reduce(|a, b| a.saturating_add(b))
-                .unwrap_or(near_primitives::gas::Gas::ZERO)
-                .as_gas();
+                .filter_map(|a| match a {
+                    Action::FunctionCall(call) | Action::FunctionCallWeight { call, .. } => {
+                        Some(call.gas.as_gas())
+                    }
+                    _ => None,
+                })
+                .sum::<u64>();
```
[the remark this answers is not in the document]
peer2f00l: Fixed in `37e859fa` — reflexive gas now adds a configurable flat per-action overhead (`--ua-reflexive-action-overhead-tgas`, default 1 Tgas) for non-call actions; reconciled against actual `tokens_burnt`.

gateway/methods-dispatch/src/account_impl.rs
```diff
@@ -45,19 +43,50 @@ impl<C: HasNearClient> DispatchRead<account::Get, C> for Dispatch {
...
+                    method_names: function_call.method_names,
+                }
+            }
+        };
+
+        Ok(account::GetAccessKeyResult {
+            nonce: key.nonce.0,
+            permission,
+        })
+    }
```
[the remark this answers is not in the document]
peer2f00l: Done in `43b9ea71` — extracted the mapping into `permission_view` and added dispatch tests for `FullAccess`, `FunctionCall` (asserting `allowance`/`receiver_id`/`method_names`), and the invalid-`receiver_id` rejection.

gateway/methods-spec/src/account.rs
```diff
@@ -24,6 +24,37 @@ pub struct GetResult {
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
+pub enum AccessKeyPermission {
+    /// Full access to the account.
+    FullAccess,
+    /// Restricted to function calls against `receiver_id` (optionally limited to
+    /// `method_names` and a remaining `allowance`).
+    FunctionCall {
+        allowance: Option<NearToken>,
+        receiver_id: String,
+        method_names: Vec<String>,
```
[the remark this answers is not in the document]
peer2f00l: Done in `43b9ea71` — `FunctionCall` now uses `AccountId` for `receiver_id` and `Vec<ContractMethodName>` for `method_names`; the dispatch parses/wraps from near_api's strings (parse failure surfaces as a gateway error).

service/relayer/src/client/database.rs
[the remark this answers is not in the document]
peer2f00l: Effectively addressed by `3f599516`: `set_pending_transaction` now claims the slot first (so a contended account can't leave an orphaned `pending` row), and the broom now `Release`s an operation that reached terminal failure with no tx hash (rejected pre-submission) instead of deferring it forever. Closing.

gateway/types/src/primitive.rs
```diff
@@ -168,6 +168,73 @@ impl JsonSchema for Base64Bytes {
...
+        Ok(Self(
+            near_api_types::transaction::delegate_action::SignedDelegateAction::try_from_slice(
+                bytes,
+            )?,
+        ))
+    }
+
+    #[must_use]
+    pub fn into_inner(self) -> near_api_types::transaction::delegate_action::SignedDelegateAction {
+        self.0
```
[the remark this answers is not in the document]
peer2f00l: Addressed in `b7b8248f` by making the type honest rather than adding off-chain crypto. The newtype guarantees *well-formedness* (borsh-decodable at the boundary), and the doc now says exactly that — authenticity (signature/expiry) is verified **on-chain** by the receiver contract per NEP-366. A forged/expired delegate action isn't a security hole here: the on-chain verification rejects it, so nothing unauthorized executes; the only cost is wasted relayer gas. Off-chain signature pre-checking purely to avoid that gas is a separate optimization (not a correctness/security fix) and out of scope for this PR.

service/relayer/src/client/database.rs
```diff
@@ -231,293 +216,146 @@ WHERE
...
-    pending_transaction_hash
+    account_id
 "#,
-            transaction_hash.to_string(),
             account_id.as_str(),
-            Decimal::from(allowance_lock_gas.as_yoctonear()),
-            Decimal::from(allowance_lock_inner.as_yoctonear()),
+            operation_key,
+            Decimal::from(inner_spend.as_yoctonear()),
         )
```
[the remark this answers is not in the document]
peer2f00l: The `account` table has a `BEFORE UPDATE` trigger (`updated_at_trigger`, migration `20250815064726`) that sets `updated_at = NOW()` on every update, so `lock_pending`'s UPDATE already refreshes the lock timestamp — a freshly-locked charge is not broom-eligible until it ages past `min_age`. Added a regression test (`lock_pending_refreshes_charge_age`, commit 93bf35e5) that backdates a stale account, locks a charge, and asserts `get_pending_charges` does not return it — so the invariant survives even if the trigger is later removed.

gateway/core/src/operation_driver.rs
```diff
@@ -82,27 +98,80 @@ impl OperationDriver {
...
+        plan: OperationPlan,
+    ) -> GatewayResult<WriteOperationResult> {
+        let mut operation = operation;
+        operation.planned = true;
+
+        if plan.steps.is_empty() {
+            // No-op: persist the terminal, step-less operation for idempotency.
+            self.store.save_operation(operation.clone()).await?;
+            return Ok(operation.record().into());
+        }
```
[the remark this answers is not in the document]
peer2f00l: Fixed in 4a11bdbc — the no-op save is now best-effort (logs a warn and returns the outcome), matching the sibling terminal save below, so a completed no-op does not surface a transient store error to the caller.

service/relayer/src/client/database.rs
```diff
@@ -231,293 +232,191 @@ WHERE
...
-"#,
-                result.transaction_hash,
-                account_id.as_str(),
-            )
-            .execute(&mut *tx)
-            .await?;
+        match operation.status {
+            OperationStatus::Succeeded => {
+                self.settle(account_id, operation_key, operation.tokens_burnt(), true)
+                    .await?;
```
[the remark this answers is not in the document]
peer2f00l: Fixed in 4a11bdbc — `resolve_charge`'s Succeeded branch now bills `inner_spend` only when `final_outcome().is_some()`, so a planned no-op success clears the slot without charging the deposit it never attached. Added `resolve_charge_no_op_success_bills_nothing` and updated the success test to use an outcome-bearing operation.

(For the record, there was no active overcharge: the only caller that locks a non-zero `inner_spend` is `storage_deposit_top_up` → `storage::Deposit`, which always emits exactly one step and is guarded against zero deposits, so a no-op success only ever carried `inner_spend = 0`. The guard makes the no-overcharge guarantee local to settlement rather than dependent on call-site discipline.)

gateway/core/src/lib.rs
```diff
@@ -63,5 +63,11 @@ pub trait OperationStore: Send + Sync {
 
     async fn save_operation(&self, operation: StoredOperation) -> GatewayResult<()>;
 
+    /// Remove a *reservation*: an operation that never executed anything (a plan
+    /// that was a no-op, or one abandoned because planning failed). Operations
+    /// with any persisted step are left untouched, so this cannot destroy the
+    /// record of work that reached the chain.
+    async fn delete_reservation(&self, operation_id: &OperationId) -> GatewayResult<()>;
```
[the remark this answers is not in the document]
peer2f00l: Fixed in 4a11bdbc — the doc no longer classifies a planned no-op as a reservation. It now states a no-op is a terminal success retained for idempotency, and that the reservation-only precondition (planning abandoned / no persisted step) is enforced by every store impl.

gateway/core/src/operation_driver.rs
```diff
@@ -82,27 +98,80 @@ impl OperationDriver {
...
+                // error), nothing was submitted and the store still holds only
+                // the bare reservation. Drop it so accounting can release the
+                // charge now instead of waiting for startup recovery.
+                if let Ok(Some(stored)) = self.store.get_by_id(&operation_id).await {
+                    if stored.is_reservation() {
+                        let _ = self.store.delete_reservation(&operation_id).await;
+                    }
+                }
+                return Err(error);
+            }
```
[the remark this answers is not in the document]
peer2f00l: Fixed in 4a11bdbc — the post-execution reservation cleanup now logs a warn on both the `get_by_id` lookup failure and the `delete_reservation` failure, matching the planning-error path, so a leaked reservation is observable rather than silently swallowed.

gateway/store/tests/driver_recovery.rs
```diff
@@ -0,0 +1,533 @@
...
+        .resume_incomplete_operations()
+        .await
+        .unwrap();
+
+    let op = store.get_by_id(&op.id).await.unwrap().unwrap();
+    assert!(matches!(
+        op.current_step,
+        Some(CurrentStep::Submitted { .. })
+    ));
+}
```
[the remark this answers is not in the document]
peer2f00l: Done in 4a11bdbc — the aged/fresh pairs are collapsed into two `#[rstest]`-parameterized tests: `reconcile_resolves_unknown_transaction_by_age` and `startup_recovery_resolves_unknown_submitted_step_by_age` (each with `case_1_aged` / `case_2_fresh`).

gateway/core/src/lib.rs
```diff
@@ -63,5 +63,10 @@ pub trait OperationStore: Send + Sync {
 
     async fn save_operation(&self, operation: StoredOperation) -> GatewayResult<()>;
 
+    /// Remove an operation (and its steps) entirely. Used for operations that
+    /// never executed anything: a reservation whose plan turned out to be a
+    /// no-op, or one abandoned because planning failed.
+    async fn delete_operation(&self, operation_id: &OperationId) -> GatewayResult<()>;
```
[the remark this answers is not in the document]
peer2f00l: Superseded by the corrected doc (see `lib.rs` at the current line / commit 4a11bdbc). `delete_reservation`'s reservation-only precondition is enforced by every impl (postgres `WHERE planned_at IS NULL`, memory `is_reservation()` guard, with a memory test), and the doc now states this rather than merely documenting the invariant. Resolving.

PR #519, opened 2026-07-22
feat(proxy-oracle): standardize contract self-upgrade (ENG-481)

peer2f00l:
Standardizes NEAR contract self-upgrade across the proxy-oracle, its governance contract, and the pyth-lazer adapter.

## What

- **`UpgradeSource` + `deploy_and_migrate` primitive** (`common/src/upgrade.rs`): one shared way to atomically deploy new code and run `migrate` in a single receipt (a failed migration reverts the deploy). Two sources — `Code` (raw blob) and `GlobalHash` (immutable, content-addressed global contract). Global-*by-account-id* is deliberately excluded: it is a live delegation to the publisher that would bypass governance and brick versioned state.
- **Wire/borsh stability**: `Code` is untagged in JSON (bare base64, matching the pre-`UpgradeSource` wire); borsh discriminants are pinned explicitly (`use_discriminant`) so persisted governance proposals stay stable across future variants.
- **Governance `SelfUpgrade` action** + `self_upgrade` TTL on the proxy-oracle governance contract, and a **v0→v1 versioned-state migration** that reshapes pending `AdminUpgrade` proposal bodies into the new `Operation` layout.
- **Framework migrate guard** (`impl_versioned_state!`): stored-vs-target version — never `migrate_args` — decides whether to migrate, and a post-run re-assert makes a multi-version-skip deploy revert atomically instead of stranding state.
- **Events**: proposal events emit a minimal `{id, kind}` (was the full proposal, incl. the wasm blob — over the 16KB log limit); pyth-lazer's `Upgraded` emits a bounded `UpgradeSummary` and is honestly framed as *scheduled*, not completed.
- **Manager CLI**: `admin-upgrade` / `self-upgrade` proposal builders sharing a typed `UpgradeArgs`.

## Testing

- Unit coverage for the untagged/tagged JSON round-trips, golden borsh discriminant bytes, the v0→v1 migration (ttls + pending-proposal reshape), and the migrate no-op/guard decision.
- Sandbox `upgrade_ordering` test: upgrades both contracts from their **real pinned mainnet blobs** to this branch, in **both orders**, seeding a real pending proposal so the on-chain reshape is exercised, and probing each upgraded contract with a domain view — proving no bricking.

Adversarial review ran to zero findings; `/simplify` applied.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/519)


common/src/upgrade.rs
```diff
@@ -0,0 +1,144 @@
...
+        }
+    }
+
+    /// Atomically deploy the new code and run `migrate_method` on it in a single receipt: a failed
+    /// migration reverts the code change too. The deploy always targets the current account, so this
+    /// is a self-upgrade primitive; the caller owns access control.
+    pub fn deploy_and_migrate(
+        self,
+        migrate_method: impl Into<String>,
+        migrate_args: Base64VecU8,
```
carrion256, line 64: Is this type not limiting calldata by ~37% rather than raw byte array?

common/src/upgrade.rs
```diff
@@ -0,0 +1,144 @@
...
+/// discriminant without disturbing the stored ones.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[near(serializers = [json, borsh(use_discriminant = true)])]
+#[repr(u8)]
+pub enum UpgradeSource {
+    /// A NEAR global contract referenced by its (immutable) code hash.
+    GlobalHash(Base58CryptoHash) = 1,
+    /// A raw WASM blob deployed onto the account.
+    #[serde(untagged)]
+    Code(Base64VecU8) = 0,
```
carrion256, line 29: Is this type not limiting calldata by ~37% rather than raw byte array? Ignoring *args, selecting base58 here reduces max contract blob. 

Also, if i was being particularly anal about this I'd mention we dont even need to decode it at all and just take the first frame as the blob.

contract/proxy-oracle/near/governance-contract/src/state.rs
```diff
@@ -0,0 +1,434 @@
...
+                        Proposal {
+                            operation: Operation::from(old.operation),
+                            created_at: old.created_at,
+                            ttl: old.ttl,
+                            created_by: old.created_by,
+                        },
+                    ))
+                })
+                .collect();
+            drop(old_proposals);
```
carrion256, line 284: Why are we smarter than the borrow checker here?

carrion256 (commented):
Really cool actually. Couple Qs

common/src/upgrade.rs
peer2f00l: Depends. `Vec<u8>` is serialized as `[1, 2, 3...]`. However, we could use Borsh, though that hurts inspectability on block explorers. But the migration arguments are usually pretty simple (usually similar in size to the arguments sent to a `new` initialization constructor function call).

common/src/upgrade.rs
peer2f00l: Yes, this is, but the only realistic way to reduce the size would be to start serializing everything with Borsh.

This should not really matter though, because we should be moving to using global contract hashes for the vast majority of contract deployments and upgrades moving forward.

contract/proxy-oracle/near/governance-contract/src/state.rs
peer2f00l: We want to run the `Drop` implementation here (we could equivalently call `.flush()`, I suppose) to actually clear the contract storage.

contract/proxy-oracle/near/contract/tests/proxy_oracle.rs
```diff
@@ -152,6 +156,44 @@ pub fn admin_upgrade_creates_one_self_receipt_with_deploy_then_migrate() {
...
+            method_name,
+            prepaid_gas,
+            ..
+        } => {
+            assert_eq!(method_name, b"migrate");
+            assert_eq!(*prepaid_gas, Contract::GAS_FOR_MIGRATE);
+        }
+        action => panic!("expected migrate call second, got {action:?}"),
+    }
+}
```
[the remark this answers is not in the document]
peer2f00l: Done in ee4635ef — the global-hash migrate action now also asserts `args == br#"{"from_version":"v0"}"#` and `attached_deposit == NearToken::from_yoctonear(0)`, matching the raw-blob sibling test.

contract/pyth-lazer/contract/src/lib.rs
```diff
@@ -496,25 +497,16 @@ impl Contract {
...
-        Promise::new(env::current_account_id())
-            .deploy_contract(code.0)
-            .function_call(
-                "migrate".to_string(),
-                migrate_args.0,
-                NearToken::from_yoctonear(0),
-                Self::GAS_FOR_MIGRATE,
-            )
+        code.deploy_and_migrate(MIGRATE_METHOD, migrate_args, Self::GAS_FOR_MIGRATE)
     }
```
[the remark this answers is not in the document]
peer2f00l: Good catch — fixed in ee4635ef. `admin_upgrade` now rejects an empty `Code` blob up front (`if code.is_empty_code() { env::panic_str(...) }`), mirroring the governance path's existing `EmptyUpgradeCode` guard, so no misleading `Upgraded` event is logged and no gas is spent on a doomed deploy. Added an `admin_upgrade_rejects_empty_code` should_panic test.

contract/pyth-lazer/contract/tests/test.rs
```diff
@@ -1027,19 +1039,27 @@ fn admin_withdraw_emits_withdrawn() {
...
+    // `code` is a structured summary — `CodeHash(<base58 sha256>)` — not the blob itself.
     let expected_hash = near_sdk::bs58::encode(near_sdk::env::sha256(&code)).into_string();
-    assert_eq!(event["data"]["code_hash"].as_str().unwrap(), expected_hash);
-    assert_eq!(event["data"]["migrated"], expected_migrated);
+    assert_eq!(
+        event["data"]["code"]["CodeHash"].as_str().unwrap(),
+        expected_hash
+    );
+    // `migrated` is whether a migration was requested — a bounded flag, not the payload.
+    assert_eq!(event["data"]["migrated"], !migrate_args.is_empty());
```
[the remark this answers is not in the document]
peer2f00l: Done in ee4635ef. `admin_upgrade_emits_upgraded` is now cross-parameterized over source × migrate_args (`#[values(...)]`), so it covers both `UpgradeSource::Code` and `UpgradeSource::GlobalHash`. The expected event payload is derived from the typed `UpgradeSource::summary()` rather than a hand-computed hash, which exercises both the `CodeHash` and `GlobalHash` summary mappings.

contract/proxy-oracle/near/governance-contract/src/lib.rs
```diff
@@ -261,6 +290,10 @@ impl ProxyGovernanceInterface for Contract {
                     )
                     .detach();
             }
+            Operation::SelfUpgrade { code, migrate_args } => {
+                code.deploy_and_migrate(MIGRATE_METHOD, migrate_args, Self::GAS_FOR_MIGRATE)
+                    .detach();
+            }
```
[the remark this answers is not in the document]
peer2f00l: Leaving this as-is, deliberately. A failed self-migrate reverts atomically within its own receipt (the `migrate` panic rolls back the code swap), so there's no partial-brick risk — only the "proposal consumed, no upgrade applied" outcome, which is identical to how every other governance operation behaves (`AdminUpgrade`, `AdminFunctionCall` all `.detach()` without observing the outcome). Adding a bespoke success/failure callback for `SelfUpgrade` alone would make the dispatch inconsistent for a nitpick-level concern. If we want failed-detached-op alerting, that's better done uniformly across all operations in a follow-up rather than special-casing this one variant here.

contract/pyth-lazer/contract/tests/test.rs
```diff
@@ -1027,19 +1051,28 @@ fn admin_withdraw_emits_withdrawn() {
...
-    assert_eq!(event["data"]["code_hash"].as_str().unwrap(), expected_hash);
-    assert_eq!(event["data"]["migrated"], expected_migrated);
+    // `code` is a structured summary (a hash), never the blob — exactly `UpgradeSource::summary()`,
+    // covering both the `CodeHash` and `GlobalHash` mappings.
+    assert_eq!(
+        event["data"]["code"],
+        near_sdk::serde_json::to_value(code.summary()).unwrap()
+    );
+    // `migrated` is whether a migration was requested — a bounded flag, not the payload.
+    assert_eq!(event["data"]["migrated"], !migrate_args.is_empty());
```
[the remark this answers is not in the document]
peer2f00l: Done in e3a8e72c — `admin_upgrade_emits_upgraded` now inspects the scheduled receipt in addition to the NEP-297 event, for both variants. It asserts two actions: `actions[0]` is `MockAction::DeployContract` for `Code` / `MockAction::UseGlobalContract` for `GlobalHash`, and `actions[1]` is the `migrate` `FunctionCallWeight` with `args == migrate_args`, `attached_deposit == 0`, and `prepaid_gas == Contract::GAS_FOR_MIGRATE`. That locks the adapter-specific deploy+migrate wiring (gas constant, method name, deposit), which previously only the shared `deploy_and_migrate` unit tests covered.

contract/pyth-lazer/contract/tests/test.rs
```diff
@@ -1027,19 +1051,57 @@ fn admin_withdraw_emits_withdrawn() {
...
+    // wiring, not just the event.
+    let receipts = get_created_receipts();
+    assert_eq!(receipts.len(), 1);
+    let actions = &receipts[0].actions;
+    assert_eq!(actions.len(), 2);
+    match &code {
+        UpgradeSource::Code(_) => assert!(matches!(actions[0], MockAction::DeployContract { .. })),
+        UpgradeSource::GlobalHash(_) => {
+            assert!(matches!(actions[0], MockAction::UseGlobalContract { .. }));
+        }
```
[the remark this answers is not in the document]
peer2f00l: Done in 25eeb21b — `admin_upgrade_emits_upgraded` now asserts the forwarded payload, not just the action variant: for `Code` it destructures `DeployContract { code, .. }` and compares the bytes to the source blob; for `GlobalHash` it destructures `UseGlobalContract { contract_id: GlobalContractIdentifier::CodeHash(h), .. }` and compares the hash to the source. Pulled in `near-primitives` (already the workspace-pinned 0.34.7, matching near-sdk) as a dev-dep to name the identifier type. A regression forwarding a different source to `deploy_and_migrate` now fails.

common/src/versioned_state/macros.rs
```diff
@@ -29,16 +29,41 @@ macro_rules! impl_versioned_state {
...
+                $crate::versioned_state::Migrator::run(args);
+
+                // One `Migrator::run` advances a single version step; re-assert we reached the target
+                // so a deploy several versions ahead of stored reverts atomically instead of leaving
+                // state stranded between versions.
+                ::near_sdk::require!(
+                    !<$current_state as $crate::versioned_state::StateVersion>::needs_migration()
+                        .unwrap_or_else(|e| env::panic_str(&e.to_string())),
+                    "state migration incomplete: stored version is still behind target",
+                );
```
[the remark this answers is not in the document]
peer2f00l: Created follow-up issue: https://linear.app/templar-protocol/issue/ENG-517/migrate-methods-support-running-n-consecutive-migrations-in-one

peer2f00l:
This follow up should be resolved before deploying new contract versions: https://linear.app/templar-protocol/issue/ENG-517/migrate-methods-support-running-n-consecutive-migrations-in-one

common/src/upgrade.rs
carrion256: What is the issue with doing the proper way of just passing a Vec<u8> rather than intentionally limit every contract to ~63% size limit. I dont see every contract becoming a global contract just because, otherwise we shouldnt even support this type in the first place.

contract/proxy-oracle/near/governance-contract/src/state.rs
carrion256: The borrow checker calls drop when the variable goes out of scope. Why do we need to try to call it?

PR #520, opened 2026-07-22
ci: share Rust caches by build class and save on failure

peer2f00l:
## Problem

CI Rust compile times run ~25 min, with `near-integration-tests` timing out at 40 min. The cache config in `.github/actions/rust/action.yml` is the cause:

- **Per-job cache keys.** `shared-key: rust-${os}-${github.job}` gives every job its own multi-GB Rust cache. Across all workflows that's ~11 near-identical caches — well past GitHub's **10 GB per-repo** limit, which triggers LRU eviction. The least-frequently-run jobs (NEAR integration, coverage — both change-gated) are the first evicted, so they rebuild from cold nearly every run.
- **`cache-on-failure: false`.** A job slow enough to hit its `timeout-minutes` is cancelled before rust-cache's post-step saves, so it can *never* seed a warm cache: cold build → timeout → still cold. Self-reinforcing.

## Change

- Add a `cache-key` build-class input to the rust composite action, falling back to the job id when unset.
- Collapse the five host-target jobs (`code-linter`, `binary-build`, `fast-tests`, `near-integration-tests`, `artifact-drift-check`) plus `gas-report` onto one shared `host` cache. `coverage` (instrumented), `soroban` / `feature-matrix` (wasm / no_std) keep private keys via the fallback — their artifacts don't share.
- Set `cache-on-failure: true` so a timing-out job still seeds its cache. Safe: CI runs `CARGO_INCREMENTAL=0`, so there's no incremental state to corrupt.

Net: distinct Rust caches drop ~11 → ~6, keeping the total under 10 GB so gated jobs stop getting evicted into cold rebuilds.

## Validation

Reasoned from config, not from cache-hit logs — the real proof is this PR's own runs. Editing `test.yml` triggers the NEAR + soroban gates, so the suite exercises the change. Watch `near-integration-tests` cache restore/save and wall-clock across the first couple of runs (the very first run is expected to still be cold-ish while it re-seeds the new `host` key).

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/520)

### DIFF
--- a/.github/actions/free-disk-space/action.yml
+++ b/.github/actions/free-disk-space/action.yml
@@ -1,0 +1,31 @@
+name: Free disk space
+description: Reclaim runner disk space by removing preinstalled toolchains we never use.
+inputs:
+  aggressive:
+    description: >-
+      Also purge large apt packages, boost, and the agent tools cache. Slower;
+      use only for disk-heavy jobs (e.g. coverage) that otherwise run out.
+    required: false
+    default: "false"
+runs:
+  using: composite
+  steps:
+    - name: Free disk space
+      shell: bash
+      env:
+        AGGRESSIVE: ${{ inputs.aggressive }}
+      run: |
+        set -euo pipefail
+        echo "Disk before:"; df -h /
+        sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /opt/hostedtoolcache/CodeQL
+        sudo docker image prune --all --force || true
+        if [ "$AGGRESSIVE" = "true" ]; then
+          sudo rm -rf /usr/local/share/boost "${AGENT_TOOLSDIRECTORY:-}"
+          sudo apt-get remove -y '^aspnetcore-.*' '^dotnet-.*' '^llvm-.*' 'php.*' \
+            '^mongodb-.*' '^mysql-.*' azure-cli google-chrome-stable firefox \
+            powershell mono-devel || true
+          sudo apt-get autoremove -y || true
+          sudo docker system prune --all --volumes --force || true
+        fi
+        sudo apt-get clean
+        echo "Disk after:"; df -h /
--- a/.github/actions/rust/action.yml
+++ b/.github/actions/rust/action.yml
@@ -1,5 +1,5 @@
 name: Setup Rust
-description: Install Rust version and setup caching
+description: Install the Rust toolchain and wire up compilation + dependency caching.
 inputs:
   targets:
     description: Rust targets to install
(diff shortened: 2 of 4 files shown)


peer2f00l:
## Status: caching works; NEAR now red on a **pre-existing dev regression** (not this change)

### What this PR does (working as intended)
- Consolidated per-job Rust caches by build class (`host`, `contracts`) + `cache-on-failure: true`. Cuts the repo from ~11 caches to ~6, under GitHub's 10 GB cap, so infrequent jobs stop getting evicted into cold rebuilds.
- **`near-integration-tests` no longer times out**: it went from cancelled-at-40m to completing in ~35m (host-side compile now cached; the full sandbox suite runs to the end).

### Why NEAR is red anyway
Completing the suite for the first time **unmasked 2 pre-existing failures on `dev`**:
- `universal-account::migration::from_0_2_0`
- `universal-account::migration::migrate_accepts_legacy_direct_payload`

Root cause: **#519 (ENG-481)** added an all-or-nothing migration guard in `common/src/versioned_state/macros.rs` (`require!(!needs_migration(), "state migration incomplete: stored version is still behind target")`), but the universal-account migration tests still exercise **stepwise** migration (`migrate(V0)` 0→1 while target is 2) and were not updated. The guard reverts the partial migrate, so `assert_success()` fails.

This is broken on `dev` itself — invisible only because the NEAR job has been timing out (the ENG-508 problem this PR fixes). `feature/eng-517-...` (run N consecutive migrations in one call) looks like the intended fix.

### Blocker
This PR can't go green until dev's migration tests are fixed (ENG-517 / a test update). That's a contract concern owned by ENG-481/ENG-517, **not** this CI change. Will rebase onto current `dev` once that lands.

### Known follow-up (not blocking)
The wasm contract prebuild is still uncached (`near-sdk` recompiles ~20×) because rust-cache doesn't preserve cargo-near's build dir — NEAR fits in 40m but with ~5m headroom. Candidate: a shared/persistent `CARGO_TARGET_DIR` for the prebuild. Tracked separately.

peer2f00l:
## Final state — node suite <40 min, caching fixed, workflows cleaned up

Reworked down to a small diff (**+78/−78, 4 files**) after measuring several approaches and reverting the ones that didn't pay off.

### Root cause of the 40 min timeout
`near-integration-tests` had `cache-on-failure: false`, so when it timed out the post-step never ran and it **never seeded a cache** — cold → timeout → still cold, permanently. Plus the cold prebuild was ~24 min on its own.

### What shipped
- **`cache-on-failure: true`** — the node-integration cache (1.56 GB) now saves even when the job fails/times out. Verified: warm rerun restored it `full match: true`.
- **`--no-abi` on the test prebuild** — cargo-near ran a separate ABI-generation build per contract that nothing in the suite reads (tests only load/deploy the `.wasm`). Measured ~7.3 min / 35% of the contract compile. Release blobs keep their ABI.
- **Cleanup** — 4 copy-pasted "free disk space" blocks → one `free-disk-space` composite action; repeated env vars → one workflow-level `env` block.

### Measured results (NEAR Integration)
| run | prebuild | total |
|---|---|---|
| cold, original (+ABI) | ~24 min | **timed out at 40** |
| cold, `--no-abi` | 18 min | 39 min |
| **warm, `--no-abi`** | **9 min** | **30 min** |

Caching demonstrably works: cold run seeds all per-job caches (including the failing NEAR job), warm rerun restores them and cuts the prebuild in half.

### Tried and reverted (with data)
- **sccache**: ~8% on the node suite — cdylib contracts + proc-macros don't cache, and the GHA backend threw ~1000 write-errors (throttling). Would need an external cache backend.
- **Sharding**: works on wall-clock but triples billed runner-minutes (the prebuild can't be sharded), so it's a net cost increase.
- **Cache grouping / cache-factory**: grouping poisons clippy-vs-build caches; restore-only-on-PR gave PRs zero cache benefit.

### Still red (out of scope)
The two `universal-account` migration tests fail on the #519/ENG-517 guard-vs-stepwise conflict — fixed by the ENG-517 branch, not here. Every other job is green.

PR #541, opened 2026-07-30
feat(manager): market export, the deployment validation gate (ENG-540)

peer2f00l:
Sub-PR 3 of 10 for ENG-537. Closes ENG-540. Targets the integration branch (no CI here; the master PR #540 is the gate).

## The gate

`MarketSpec::from_deployed` inverts `into_market_configuration`. The round-trip test runs **offline** against the checked-in deployment configs and asserts the exported spec re-derives to exactly the deployed `MarketConfiguration` and both proxies.

It also asserts the rendered TOML reloads identically — TOML serialization is order-sensitive about tables, so "it serializes" was an assumption worth testing rather than trusting.

## Coverage is 2 of 18, deliberately

Per your call: a spec expresses only proxy-oracle deployments. Thirteen alpha markets read `pyth-oracle.near` directly, two use a proxy named after the asset pair rather than the market (`proxy-oracle-cetes-usdc` for market `ixlmcetes-ixlmusdc`), one uses the LST oracle.

Those are **refused, not approximated** — a spec that re-derives to a different oracle account is a silent wrong answer. `refuses_a_market_it_cannot_express` asserts the error names both the deployed oracle and what a spec would derive instead.

Stated in the module docs, the test docs, and here, because a gate that quietly covers 11% reads like one that covers everything.

## Notable

- `symbol` is left unset by export rather than invented. A plausible-looking ticker would give ENG-543's reference check a guess to verify.
- Account-id derivation moved to free functions in `spec` (`oracle_account_id`, `governance_account_id`). Export needs them before it has a spec to call methods on, and a second copy of the naming convention is the duplication this epic exists to remove.
- Governance admin is a required flag: no read method exposes it, so it cannot be recovered from chain state. Version keys *are* recovered, from the registry deployment records.

## Verification

`cargo fmt` · `cargo check --workspace` clean · `cargo clippy -p templar-manager --tests -- -D warnings` clean · `just test-fast -p templar-manager` 163/163.

`/simplify` ran before this landed and produced five findings, all applied — notably that the read chain sat in `commands/`, breaking the layering `dispatch/mod.rs` documents (commands stay pure; multi-step IO lives in `dispatch/`). It now lives in `dispatch/export.rs`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/541)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -253,6 +253,18 @@ jobs:
         # is the one thing that catches it.
         run: cargo build --workspace --bins
 
+      # `tmplrmgr`'s `ledger` feature is opt-in so the default workspace build
+      # needs no system libraries — hidapi wants libudev, and cargo unifies
+      # features across a workspace build, so enabling it by default would put
+      # that dependency on all 22 crates. Compiling it here, in the one job that
+      # already installs nothing else, is what stops an opt-in feature from
+      # rotting unnoticed.
+      - name: Install Ledger build dependencies
+        run: sudo apt-get update && sudo apt-get install -y libudev-dev
+
+      - name: Check the ledger-enabled signer
+        run: cargo check -p templar-manager --features ledger --all-targets
+
   soroban:
     name: Soroban
     runs-on: warp-ubuntu-latest-x64-4x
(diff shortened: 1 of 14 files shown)


peer2f00l:
@coderabbitai review

PR #542, opened 2026-07-30
feat(manager): on-chain preflight checks (ENG-541)

peer2f00l:
Sub-PR 4 of 10 for ENG-537. Closes ENG-541. Targets the integration branch; master PR #540 is the gate.

## What

Read-only checks before any transaction: `asset.exists.*`, `asset.decimals.*`, `oracle.source.*`, `registry.version.*`, `account.exists.*`. `spec check` gains `--offline` (no network, CI-usable) and `--accept-decimals-mismatch`.

## Verified against mainnet

```
asset.decimals.collateral   passed   6, matching token metadata
asset.decimals.borrow       passed   7 declared; token publishes no metadata, so this is unverified
oracle.source.collateral.0  passed   lazer feed 14 on pyth-lazer.templar-alpha.near, currently carrying a price
registry.version.market     passed   v1.3.0
config.validate             passed   MarketConfiguration::validate
```

The borrow line is the bridged-asset case that motivated the `decimals` override: its NEP-245 token id names no NEP-141 account, so nothing on chain answers. Reported as **unverified**, not as confirmed.

## Three things worth reviewing

**Absence of a price is not evidence of an unsupported feed.** An adapter carries a feed once someone pushes one — Pyth Lazer verifies any feed it signs. A market deployed for the first time may name a feed that has never appeared on this adapter, which says nothing about support. So the adapter must exist (hard fail), but a missing price is reported as `Skipped` with the reason. Failing on it would block exactly the new-market case this pipeline exists for.

**Resolved decimals are written back into the spec before the offline checks run.** Otherwise `config.validate` reports itself skipped for want of decimals the same invocation just read off the chain — the two passes didn't compose.

**`exists` distinguishes `AccountNotFound` from other read failures.** `.is_ok()` would report a timed-out RPC as "this account does not exist", sending an operator to create an account that is already there. `GatewayError::AccountNotFound` already exists for this and is used the same way in `gateway/core/src/contract_kind.rs`.

## Verification

`cargo fmt` · `cargo clippy -p templar-manager --tests -- -D warnings` clean · `just test-fast -p templar-manager` 171/171 · mainnet smoke (read-only).

The decimals matrix is a pure function with 6 rstest cases, so the whole thing is testable without a network; the IO around it is one view call.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/542)

### DIFF
--- a/tools/manager/src/commands/spec.rs
+++ b/tools/manager/src/commands/spec.rs
@@ -1,16 +1,15 @@
-//! Local operations on a deployment spec. Nothing here touches the network —
-//! that is the property the offline checks exist to preserve.
+//! Local operations on a deployment spec. Arguments only — the preflight that
+//! fulfils `check` lives in [`crate::dispatch::preflight`], since it reads the
+//! chain.
 
 use std::path::PathBuf;
 
 use clap::{Args, Subcommand};
 
-use crate::spec::{check, extends};
-
 #[derive(Subcommand, Debug)]
 #[command(rename_all = "kebab-case")]
 pub enum SpecNs {
-    /// Resolve a spec's `extends` chain and run every offline check.
+    /// Resolve a spec's `extends` chain and run its preflight checks.
     Check(Check),
     /// Emit the spec's JSON Schema, for editor completion and validation.
     PrintSchema,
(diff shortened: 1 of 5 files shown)


peer2f00l:
@coderabbitai review

PR #543, opened 2026-07-30
fix: address the review-bot findings I merged past (ENG-538..541)

peer2f00l:
Not a sub-issue. This corrects a process failure: I merged PRs #538, #539, #541 and #542 without reading the codex and CodeRabbit reviews on them. `gh pr view --comments` shows only issue comments — the findings live in `reviews` and in the inline threads (`gh api .../pulls/N/comments`). Nine were still live.

## P1

**`market export` defaulted `ttl_default = 0s`.** For a governance contract with a real timelock, redeploying the exported spec would have **removed** it. The TTL is now read from `proxyOracleGovernance.getOperationTtl` across every operation kind; a uniform value is recovered, and a per-kind set is refused since `GovernanceSpec` carries one TTL. The `--governance-ttl` flag is gone — it was a guess where a read was available.

## Silent-corruption class

**Durations truncated.** The chain stores `price_maximum_age` in whole seconds and `time_chunk` in whole milliseconds. `price_maximum_age = "1500ms"` meant 1s to the market while the proxy still enforced 1.5s; `time_chunk = "500us"` collapsed to a **zero-length chunk**. Now rejected, with three rstest cases.

**`--network` was never compared to the spec's own chain.** The CLI defaults to testnet, so `spec check` on the mainnet spec without the flag queried testnet, reported every account and version as missing, and still printed `"network":"mainnet"` — a page of confident, entirely wrong failures.

**Export could copy a soft-deleted version key.** `remove_version` clears the code but leaves the deployment record, so the spec named a version that cannot be deployed. Now refused.

## Correctness / usability

**`--sign-with secret-key` parsed with no credential**, because a flag's mere presence satisfied the requirement. `--sign-with` now names only backends that hold the key *outside* this process, so that invocation is unrepresentable rather than discouraged. (`required_if_eq` was tried first and does not fire on a defaulted value.)

**`--public-key` no longer conflicts with `--secret-key`.** An ambient `SECRET_KEY` made the documented `--sign-with keychain --public-key …` flow fail at parse time over a variable the chosen backend ignores.

**A failed read aborted the whole report** instead of recording a failed check — one flaky RPC hid every other problem in the spec.

**`account.exists.yield_static` reused one id for every recipient**, in `HashMap` order. Check ids are a contract (`--skip-check` and the plan artifact key on them), so each recipient gets its own, in a stable order.

**`symbol` became optional in ENG-540 without a schema bump**, so a schema-1 reader would reject an exported spec. Bumped to 2.

## Also

`Cargo.lock` was stale — the ledger removal dropped ~95 lines that were never committed.

## Verification

`cargo fmt` · `cargo check --workspace` clean · `cargo clippy -p templar-manager --tests -- -D warnings` clean · `just test-fast -p templar-manager` **176/176** (5 new regression tests).

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/543)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4600,19 +4600,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6fe2267d4ed49bc07b63801559be28c718ea06c4738b7a03c94df7386d2cde46"
 
 [[package]]
-name = "hidapi"
-version = "2.6.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c78dadfc12f865bc3fcac3897e64533b930737ceb9ef245c8277de98d0b010e9"
-dependencies = [
- "cc",
- "cfg-if 1.0.4",
- "libc",
- "pkg-config",
- "windows-sys 0.61.2",
-]
-
-[[package]]
 name = "hidapi-rusb"
 version = "1.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


peer2f00l:
@codex review

PR #587, opened 2026-08-07
feat(common): registry view types for sound target and version checks

peer2f00l:
First of five stacked PRs clearing the **Registry Contract** project. Merge bottom-up; this one is the base.

Adds the wire types the registry's new read views return, so `tmplrmgr` can establish two things it currently cannot: whether a target name is claimed, and whether a registered version is still deployable.

Both are enums rather than the booleans ENG-559 and ENG-560 originally specced. A version has four distinguishable states and no single boolean separates them: a `GlobalHash` version stores no code yet deploys fine, so "has code" reports it alongside a soft-deleted one — and on mainnet *every* registered version is `Normal`-mode, making that the common case rather than the edge case. `RegistryEntryView` likewise carries `Reserved`, which `Deployment` alone cannot express.

`VersionInfo::code_hash` is the registry's own `env::sha256_array` of the code, so it is a sound key for resolving a wasm from the artifact cache — unlike the digest embedded in a version key, which is a convention the registry does not enforce and which several live version keys omit entirely.

`Deployment` gains `PartialEq`/`Eq` so the view types and their gateway results can derive them.

Relates to ENG-559, ENG-560.

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/587)

### DIFF
--- a/common/src/registry.rs
+++ b/common/src/registry.rs
@@ -20,10 +20,128 @@ impl std::fmt::Display for DeployMode {
     }
 }
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 #[near(serializers = [borsh, json])]
 pub struct Deployment {
     pub version_key: String,
     pub code_hash: Base58CryptoHash,
     pub block_height: U64,
+}
+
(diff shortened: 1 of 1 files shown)


peer2f00l:
@codex review

PR #619, opened 2026-09-01
fix(soroban): report zero for empty Blend positions

carrion256:
## Summary

- treat an absent Blend supply-map entry as zero exposure
- keep the existing `MissingPosition = 5` discriminant reserved for deployed ABI compatibility
- cover empty-adapter refreshes and reads after a complete Blend exit

## Bug

Blend stores account positions in sparse maps. Before an adapter supplies for the first time, and
after it withdraws its complete position, `positions.supply` has no entry for the reserve index.
That absence means the account has zero b-tokens.

The adapter instead converted the missing entry into `AdapterError::MissingPosition`:

```rust
positions.supply.get(index).ok_or(AdapterError::MissingPosition)?
```

The vault calls `total_assets` directly during `RefreshMarkets`, so the adapter error escalated
through the cross-contract call. A correctly configured but unused market therefore could not be
refreshed, and the same liveness failure could return after a complete market exit even though the
true external exposure was zero.

Initial allocation generally still worked because the supply call creates the Blend position
before the vault reads `total_assets`. This is a refresh/reconciliation liveness bug rather than a
deposit-accounting or current loss-of-funds bug.

## Fix

Use zero as the default b-token balance when the sparse map has no reserve entry. Reserve lookup,
staleness validation, and checked b-rate arithmetic remain unchanged.

## Regression coverage

- adapter with no prior supply reports `0`
- adapter remains readable and reports `0` after withdrawing its full position
- vault `RefreshMarkets` succeeds with zero external assets before first allocation

Before the implementation change, the adapter regression failed with
`left: Err(MissingPosition), right: Ok(0)`, and the vault regression failed with contract error
`#5`.

## Verification

- `cargo +1.89.0 fmt --all --check`
- `cargo +1.89.0 test -p templar-soroban-blend-adapter -- --nocapture` — 30 passed
- `cargo +1.89.0 test -p templar-soroban-runtime -- --nocapture` — 265 passed, 1 doc test ignored
- `cargo +1.89.0 clippy -p templar-soroban-blend-adapter -p templar-soroban-runtime --all-targets --all-features -- -D warnings`
- release Soroban target compilation for `templar-soroban-blend-adapter` on `wasm32v1-none`

- - -
This change is [Reviewable](https://reviewable.io/reviews/Templar-Protocol/contracts/619)

### DIFF
--- a/contract/vault/soroban/blend-adapter/CHANGELOG.md
+++ b/contract/vault/soroban/blend-adapter/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Report zero assets when the adapter has no Blend supply position, keeping vault refreshes live
+  before first allocation and after a complete exit.
+
 ## [1.0.1](https://github.com/Templar-Protocol/contracts/compare/templar-soroban-blend-adapter-v1.0.0...templar-soroban-blend-adapter-v1.0.1) - 2026-08-03
 
 ### Added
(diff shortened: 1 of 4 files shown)


carrion256:
@coderabbitai review

carrion256:
This needs a publish hotfix

